PR #10, opened 2024-02-20
Make MAC addresses build-time configurable

wom-bat:
The Makefile will now generate a random MAC address range to use for all components.  Individual components add an offset to the base MAC address

The Makefile saves the MAC address it uses and reuses the same one if it finds a saved one, so rerunning make does not change the MAC address unless you delete the build tree.

This relies on the corresponding changes to the sDDF repo.


Ivan-Velickovic:
Going to close this since I'm about to integrate the new tooling which will do effectively the same thing.

Issue #2, opened 2023-11-29
Outstanding issues with NFS client


JE-Archer:
- [x] os.listdir() on some directories causes a crash (IP=NULL), e.g. os.listdir('content/images').
- [ ] need a proper asynchronous client library
- [x] need examples in the README on how to use it
- [x] dependency on libgcc
- [x] libnfs doesn't work with clang
- [x] libnfs doesn't build on macos due to unsupported flag
- [x] some stuff not implemented, e.g. file modes
~~- [ ] using musl instead of our own libc~~ This is it's own GitHub issue https://github.com/au-ts/lionsos/issues/48
- [ ] musl calls into gettime() to get real-time clock, but we don't have an rtc
- [x] unsure of memory lifetimes of libnfs, i.e. how robust is our heap management to long-running sessions?
- [x] unsure of socket lifetimes, i.e. do we need to do anything extra to handle disconnects/reconnects etc?
- [x] after leaving the Kitty example open for a while, ` nfs: tcp.c:331:socket_recv_callback: closing connection...` appears...
- [x] use upstream libnfs instead of our fork

PR #50, opened 2024-05-01
Preliminary work to enable webserver

JE-Archer:
This PR includes preliminary work to enable a later PR for the webserver system. It:
- improves the FS interface;
- improves the NFS client;
- improves the MP FS implementation;
- changes how the MP event loop works;
- adds a new MP module for using the FS interface asynchronously;
- improves MP and NFS TCP performance;
- updates libnfs to the latest upstream commit.


examples/kitty/src/micropython/micropython.c
```diff
@@ -70,7 +70,7 @@ STATIC bool init_nfs(void) {
 #ifndef NDEBUG
 void MP_WEAK __assert_func(const char *file, int line, const char *func, const char *expr) {
     // @ivanv: improve/fix, use printf?
```
Ivan-Velickovic, line 72: Remove this comment now this is fixed

examples/kitty/src/micropython/mpnetworkport.c
```diff
@@ -83,7 +83,7 @@ static bool notify_rx = false;
 #define LINE_INDEX(a) (LINE_START(a)>>CONFIG_L1_CACHE_LINE_SIZE_BITS)
 
 u32_t sys_now(void) {
-    return mp_hal_ticks_ms();
+    return mp_time_time_get();
```
Ivan-Velickovic, line 86:
```suggestion
    /* LWIP expects the current time in milliseconds */
    return mp_time_time_get();
```

fs/include/fs/protocol.h
```diff
@@ -42,10 +42,10 @@ union sddf_fs_message {
 };
 
 struct sddf_fs_queue {
+    uint64_t head;
+    uint64_t tail;
+    uint64_t padding[6];
```
Ivan-Velickovic, line 47:
```suggestion
    /* Add explicit padding to ensure buffer entries are cache-entry aligned. */
    uint64_t padding[6];
```

examples/kitty/src/micropython/fs_helpers.c
```diff
@@ -0,0 +1,130 @@
...
+    bool complete;
+} request_metadata[SDDF_FS_QUEUE_CAPACITY];
+
+#define NUM_BUFFERS SDDF_FS_QUEUE_CAPACITY * 4
+struct buffer_metadata {
+    bool used;
+} buffer_metadata[SDDF_FS_QUEUE_CAPACITY];
+
+int fs_request_allocate(uint64_t *request_id) {
+    for (uint64_t i = 0; i < NUM_BUFFERS; i++) {
```
Ivan-Velickovic, line 27: Use the same type for incrementing everywhere (it's different in `fs_buffer_allocate`)

Ivan-Velickovic:
The Kitty example cannot import over NFS as it looks like the newer libnfs version fails to open if the permissions do not match. The newer libnfs behaviour seems correct and instead we should be respecting modes correctly and not using read-write for everything as outlined in https://github.com/au-ts/lionsos/issues/2.

Once this is fixed in the PR and everything in Kitty still works we can merge.

examples/kitty/src/micropython/micropython.c
```diff
@@ -52,7 +53,6 @@ void await(int event_source) {
     }
     mp_blocking_events = event_source;
     co_switch(t_event);
-    assert(active_events & event_source);
```
Ivan-Velickovic, line 55: Not sure why this is necessary to be removed, should check.

Ivan-Velickovic (approved):

PR #61, opened 2024-05-20
Update to latest sDDF

Kswin01:
Update the sDDF hash, and modify the kitty system appropriately.


Ivan-Velickovic:
In the commit message for the commit that removes the ARP component, can you add a line saying that we don't need the ARP component is not necessary in the case of the Kitty example because our clients use lwIP which can handle ARP requests on its own.

examples/kitty/Makefile
```diff
@@ -195,6 +197,7 @@ NETWORK_COMPONENTS_INCLUDE := \
 	-Isrc/ethernet_config
 NETWORK_COMPONENT_CFLAGS := \
 	$(NETWORK_COMPONENTS_INCLUDE) \
+	-DNUM_NETWORK_CLIENTS=2 \
```
Ivan-Velickovic, line 200: I thought the ethernet config file already had this?

examples/kitty/src/ethernet_config/ethernet_config.h
```diff
@@ -2,9 +2,8 @@
 
 #include <sddf/network/queue.h>
 
-#define NUM_CLIENTS 3
+#define NUM_CLIENTS 2
```
Ivan-Velickovic, line 5: This should probably be renamed to `NETWORK_NUM_CLIENTS` for the latest sDDF change I think.

examples/webserver/Makefile
```diff
@@ -171,6 +174,9 @@ NETWORK_COMPONENT_CFLAGS := \
 NETWORK_COMPONENTS_LDFLAGS := -L$(BOARD_DIR)/lib -L$(SDDF)/lib -L$(LIBGCC)
 NETWORK_COMPONENTS_LIBS := -lmicrokit -Tmicrokit.ld -lc -lgcc $(BUILD_DIR)/sddf_util_cache.o $(BUILD_DIR)/sddf_util_printf.o
 
+NETWORK_COMPONENTS_LDFLAGS := -L$(BOARD_DIR)/lib -L$(LIBGCC) -L$(LIBC)
+NETWORK_COMPONENTS_LIBS := -lmicrokit -Tmicrokit.ld -lc -lgcc $(BUILD_DIR)/sddf_util_cache.o $(BUILD_DIR)/sddf_util_printf.o
+
```
Ivan-Velickovic, line 179: These are getting defined twice?

Ivan-Velickovic: Lines 174 and 175 already seem to define them

Kswin01: Also removed libgcc dependency here.

Ivan-Velickovic (approved):

PR #69, opened 2024-05-29
Kitty submodules

Kswin01:
Intialise the required submodules in the Kitty Makefile if they have not yet been initialised.

### DIFF
--- a/examples/kitty/Makefile
+++ b/examples/kitty/Makefile
@@ -284,14 +284,18 @@ directories:
 	$(shell mkdir -p $(BUILD_DIR)/lwip/src/core/ipv4)
 	$(shell mkdir -p $(BUILD_DIR)/nfs)
 
-submodules:
-	git submodule update --init
-	cd ${LIONSOS}/dep/micropython && git submodule update --init lib/micropython-lib
-
 $(IMAGE_FILE) $(REPORT_FILE): $(addprefix $(BUILD_DIR)/, $(IMAGES)) kitty.system
 	$(MICROKIT_TOOL) kitty.system --search-path $(BUILD_DIR) --board $(MICROKIT_BOARD) --config $(MICROKIT_CONFIG) -o $(IMAGE_FILE) -r $(REPORT_FILE)
 
 FORCE: ;
+
+submodules:
+	git submodule update --init $(LIONSOS)/dep/libvmm
+	git submodule update --init $(LIONSOS)/dep/libnfs
+	git submodule update --init $(LIONSOS)/dep/micropython
+	git submodule update --init $(LIONSOS)/dep/musllibc
+	git submodule update --init $(LIONSOS)/dep/sddf
+	cd ${LIONSOS}/dep/micropython && git submodule update --init lib/micropython-lib
 
 mpy-cross:
 	make -C $(LIONSOS)/dep/micropython/mpy-cross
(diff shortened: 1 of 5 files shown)


Ivan-Velickovic:
If you could make a PR to update the documentation on lionsos.org that would be great as well.

Ivan-Velickovic:
What happens if you're developing and you change the branch of a submodule, when you run `make` won't it go back to the checked in commit?

I think this is why we should have a separate `make submodules` step and `make` just does the compiling.

Ivan-Velickovic:
You can look at the webserver, I believe it does this.

Kswin01:
> What happens if you're developing and you change the branch of a submodule, when you run `make` won't it go back to the checked in commit?

That's what the git submodule status check was for, to only initialise the submodules if they haven't already been initialised, as denoted by the '-' in front of the submodule.
 
> I think this is why we should have a separate `make submodules` step and `make` just does the compiling.

Ok, best to keep it inline with the webserver as well then. I'll make those changes now.

Ivan-Velickovic:
> That's what the git submodule status check was for, to only initialise the submodules if they haven't already been initialised, as denoted by the '-' in front of the submodule.

But then if you do a `git pull` and run `make submodules` it won't update anything right?

examples/kitty/Makefile
```diff
@@ -284,6 +284,10 @@ directories:
 	$(shell mkdir -p $(BUILD_DIR)/lwip/src/core/ipv4)
 	$(shell mkdir -p $(BUILD_DIR)/nfs)
 
+submodules:
+	git submodule update --init
```
Ivan-Velickovic, line 288: This will initialise every submodule we have, whereas we want to only initialise the ones we need.

Ivan-Velickovic: (right now I think that's everything in `dep/` except `microdot`.

Kswin01: I was following what was in the LionsOS documentation, which does `git submodule update --init`. I'll reduce down to just the ones we need.

Ivan-Velickovic: Ah okay, yea that documentation is wrong then.

I guess until recently all the submodules were dependencies of Kitty but not anymore.

examples/kitty/Makefile
```diff
@@ -289,7 +289,17 @@ $(IMAGE_FILE) $(REPORT_FILE): $(addprefix $(BUILD_DIR)/, $(IMAGES)) kitty.system
 
 FORCE: ;
 
+init-submodules:
```
Ivan-Velickovic, line 292: Why not keep it consistent with the webserver example?

Kswin01: Because in the webserver example you'll have to do `make submodule`, then `make` commands separately. This way you can do just  `make submodule` and it'll build as well. Do you prefer doing it in two distinct steps?

Ivan-Velickovic: If you want to do everything you can do `make submodule all`. I prefer the two distinct steps since it's clear what's doing.

For me at least, it's not that intuitive to have `make submodules` also start compiling code.

Kswin01: Alright, I've updated that. I'll quickly update the docs PR as well.

Ivan-Velickovic (approved):
Thanks for adding this.

PR #70, opened 2024-06-03
micropython: don't use printf 'h' format specifiers

JE-Archer:
The MicroPython component uses lwIP for its socket implementation. lwIP in turn makes use of the printf provided by MicroPython. MicroPython's printf does not support the 'h' specifiers, like %hu or %hx. As such, lwIP must not call printf with these specifiers. lwIP's usage of these specifiers can be configured via the U16_F and X16_F constants in lwip_include/arch/cc.h. This commit changes those constants to use %u and %x instead of %hu and %hx.


wom-bat (approved):
Looks good.  We don't use any 16-bit platforms so this should always be safe

PR #76, opened 2024-06-26
Make FS protocol more explicit

JE-Archer:


include/lions/fs/protocol.h
```diff
@@ -31,67 +87,228 @@ enum {
...
+typedef struct fs_queue {
     uint64_t head;
     uint64_t tail;
     /* Add explicit padding to ensure buffer entries are cache-entry aligned. */
-    uint64_t padding[6];
-    union fs_message buffer[FS_QUEUE_CAPACITY];
-};
+    uint8_t padding[48];
+    fs_msg_t buffer[FS_QUEUE_CAPACITY];
+} fs_queue_t;
```
Ivan-Velickovic, line 288: Would be good to have a static assert to check that this is cache aligned.

components/fs/nfs/nfs.c
```diff
@@ -101,15 +97,19 @@ void notified(microkit_channel ch) {
         /* Nothing to do in this case */
         break;
     case CLIENT_CHANNEL:
-        if (nfs_connected) {
```
Ivan-Velickovic, line 104: `nfs_connected` is not used anymore, should it be removed?

Ivan-Velickovic (commented):
I need to keep going through the code but at a high level I don't see any issues.

We need to start documenting things like invariants, for example `process_commands` function enforces invariants that are assumed by the command handlers. Having things like this commented makes it much easier to read and reason about the code.

We also need to document this protocol, both in the design document and in `protocol.h`.

components/micropython/micropython.c
```diff
@@ -92,8 +92,11 @@ void t_mp_entrypoint(void) {
     gc_init(heap, heap + sizeof(heap));
     mp_init();
 
-    init_nfs();
     init_networking();
+    // initialisation of the filesystem utilises the event loop and the event
+    // loop unconditionally tries to process incoming network buffers; therefore
+    // the networking needs to be initialised before initialising the fs
+    init_nfs();
```
Ivan-Velickovic, line 99:
```suggestion
    /*
     * In the case of the network file system, initialisation of the filesystem utilises the event loop and the event
     * loop unconditionally tries to process incoming network buffers; therefore
     * the networking needs to be initialised before initialising the FS
     */
    init_fs();
```

Ivan-Velickovic: Maybe `init_vfs` would be better for MicroPython?

Ivan-Velickovic: Nevermind, I mis-interpreted the original comment.

include/lions/fs/protocol.h
```diff
@@ -31,67 +87,228 @@ enum {
...
 }
 
-static bool fs_queue_pop(struct fs_queue *queue, union fs_message *message) {
-    if (queue->head == __atomic_load_n(&queue->tail, __ATOMIC_ACQUIRE)) {
-        return false;
-    }
-    *message = queue->buffer[queue->head % FS_QUEUE_CAPACITY];
-    __atomic_store_n(&queue->head, queue->head + 1, __ATOMIC_RELEASE);
-    return true;
+static inline uint64_t fs_queue_size_producer(fs_queue_t *queue) {
```
Ivan-Velickovic, line 294: Is `queue_length` a bit less ambiguous that `queue_size` or is it fine how it is? sDDF isn't consistent right now, some things use length some use size.

components/fs/nfs/op.c
```diff
@@ -25,13 +25,15 @@
...
 
-#define MAX_CONCURRENT_OPS 100
+#define MAX_CONCURRENT_OPS FS_QUEUE_CAPACITY
 #define CLIENT_SHARE_SIZE 0x4000000
 
 struct fs_queue *command_queue;
 struct fs_queue *completion_queue;
 char *client_share;
 
+char path_buffer[4096][2];
```
Ivan-Velickovic, line 35:
```suggestion
char path_buffer[FS_MAX_PATH_LENGTH][2];
```
?

Although that would be 4095 instead of 4096.

Ivan-Velickovic: It should be `FS_MAX_PATH_LENGTH + 1` now that I look at `copy_path`

Ivan-Velickovic (approved):

Ivan-Velickovic (approved):

PR #81, opened 2024-07-15
sDDF compatibiity changes for LionsOS Kitty example

wom-bat:
-- Introduce serial_config.h
-- Move both config files (ethernet_config.h and serial_config.h) into
   a config directory
-- Adjust micropython glue files to use new serial interfaces -- Adjust system description file to use new serial interfaces -- Adjust nfs subsystem to use new serial interfaces -- Update to latest libvmm and sDDF

Compiles, but untested.


examples/webserver/webserver.mk
```diff
@@ -0,0 +1,132 @@
...
+	-I$(SDDF)/include \
+	-I$(CONFIG_INCLUDE)
+
+LDFLAGS := -L$(BOARD_DIR)/lib
+LIBS := -lmicrokit -Tmicrokit.ld libsddf_util_debug.a
+
+IMAGE_FILE := webserver.img
+REPORT_FILE := report.txt
+
+#.EXTRA_PREREQS := submodules
```
Ivan-Velickovic, line 53: I assume this line is not needed, so will remove it.

components/fs/nfs/nfs.h
```diff
@@ -2,8 +2,10 @@
  * Copyright 2023, UNSW
  * SPDX-License-Identifier: BSD-2-Clause
  */
+#include <sddf/serial/queue.h>
 
-
+#define SERIAL_TX_CH 0
+extern serial_queue_handle_t serial_tx_queue_handle;
```
Ivan-Velickovic, line 8: Why are we externing this now?

examples/webserver/Makefile
```diff
@@ -7,239 +7,44 @@
...
-	-I$(BOARD_DIR)/include \
-	-target $(TARGET) \
-	-DBOARD_$(MICROKIT_BOARD) \
-	-I$(SDDF)/include
-LDFLAGS := -L$(BOARD_DIR)/lib
-LIBS := -lmicrokit -Tmicrokit.ld
 
 IMAGE_FILE := $(BUILD_DIR)/webserver.img
 REPORT_FILE := $(BUILD_DIR)/report.txt
+nproc:= $(shell sh -c "if test -f /proc/cpuinfo; then grep processor /proc/cpuinfo | wc -l; else echo 2; fi")
```
Ivan-Velickovic, line 31: Hmm. Not sure about this, it should be up to the user to decide how many threads Make should start. Anyways, we can sort that out later.

examples/webserver/config/ethernet_config.h
```diff
@@ -0,0 +1,183 @@
+#pragma once
+#ifndef KITTY_ETHERNET_CONFIG_H
+#define KITTY_ETHERNET_CONFIG_H
```
Ivan-Velickovic, line 3: Do we need this? I thought `#pragma once` made it not necessary.

PR #88, opened 2024-08-08
Add QEMU support for the web server example

Ivan-Velickovic:
Allows the web server to be run on the QEMU simulator.

There are two things to note:
* the performance is not great, it still loads pages within a couple of seconds but it's significantly slower than the Odroid-C4. We're not exactly sure why right now but we're investigating.
* We use host-forwarding from port 5555 to port 80 on the system, which is the port the web server starts on. In order to get this working though, I had to specify the IP address of the MicroPython client. I'm not sure why this was necessary since the two clients (MicroPython and NFS) had different MAC addresses so it should've been fine...


examples/webserver/Makefile
```diff
@@ -19,21 +19,19 @@ endif
...
 IMAGE_FILE := $(BUILD_DIR)/webserver.img
 REPORT_FILE := $(BUILD_DIR)/report.txt
 nproc:= $(shell sh -c "if test -f /proc/cpuinfo; then grep processor /proc/cpuinfo | wc -l; else echo 2; fi")
 
 all: ${IMAGE_FILE}
 
-${IMAGE_FILE}: ${BUILD_DIR}/Makefile FORCE
-	${MAKE} -C ${BUILD_DIR} MICROKIT_SDK=${MICROKIT_SDK}  $(notdir $@) 
+qemu ${IMAGE_FILE}: ${BUILD_DIR}/Makefile FORCE
+	${MAKE} -C ${BUILD_DIR} MICROKIT_SDK=${MICROKIT_SDK}  $(notdir $@)
```
alwin-joshy, line 34: Doesn't `make qemu` usually run the image?

alwin-joshy: Nevermind, I understand

alwin-joshy (approved):

Kswin01 (approved):

PR #89, opened 2024-08-08
kitty: make vm images explicit dependencies

JE-Archer:
This removes the "make images" build step and allows make to download the images automatically when required.


Ivan-Velickovic:
I'll update the docs site now.

PR #91, opened 2024-08-08
Download submodules and images on demand

wom-bat:
instead of having to have a separate target for getting them.


examples/kitty/kitty.mk
```diff
@@ -164,16 +163,42 @@ nfs.elf: nfs/nfs.a libnfs/lib/libnfs.a musllibc/lib/libc.a
...
+	git submodule update --init $(LIONSOS)/dep/libvmm
+
+${LIBNFS}/nfs:
+	git submodule update --init $(LIONSOS)/dep/libnfs
+
+$(LIONSOS)/dep/micropython/py/mkenv.mk ${LIONSOS}/dep/micropython/mpy-cross:	
+	git submodule update --init $(LIONSOS)/dep/micropython
+	cd ${LIONSOS}/dep/micropython && git submodule update --init lib/micropython-lib
+
+	git submodule update --init $(LIONSOS)/dep/musllibc
```
Ivan-Velickovic, line 199: This one should be for libnfs.

Ivan-Velickovic:
I think this should work, I made the submodules a separate target because I wanted people to develop/change the submodules and be able to test the Kitty example without having all the submodules reset by the Makefile. I think these changes should work in that case though.

Can you make the same submodules changes for the web server as well?

wom-bat:
> I think this should work, I made the submodules a separate target because I wanted people to develop/change the submodules and be able to test the Kitty example without having all the submodules reset by the Makefile. I think these changes should work in that case though.
> 
> Can you make the same submodules changes for the web server as well?

Want to get this one working on the CI first.

wom-bat:
> Want to get this one working on the CI first.

Seems to be a race with -j2 --- works with -j1.  I think I've fixed it, testing now ...

examples/kitty/kitty.mk
```diff
@@ -164,16 +160,45 @@ nfs.elf: nfs/nfs.a libnfs/lib/libnfs.a musllibc/lib/libc.a
...
+${LINUX}:
+	curl -L ${LIONSOS_DOWNLOADS}/$(KITTY_GRAPHICS_VM_LINUX) -o $@
+
+${INITRD}:
+	curl -L ${LIONSOS_DOWNLOADS}/$(KITTY_GRAPHICS_VM_ROOTFS) -o $@
+
+# Because we include the *.mk files they are an implicit dependency
+# for the build.  These rules instantiate the submodules that
+# include the makefiles.
+${LIBVMM_DIR}/vmm.mk:
```
Ivan-Velickovic, line 186: These check for the *existence* of these files not if the files have changed right? E.g if you pulled a new version of libvmm that changed `vmm.mk` would it update the submodule?

wom-bat:
> > Can you make the same submodules changes for the web server as well?
Done

PR #100, opened 2024-08-20
Add missing error check in posix TCP handling

Kswin01:
This PR properly enables TCP keep-alive for lwip. Additionally we add proper error checking to the posix layer, where before we were not checking the return value of tcp socket create and assigning bad socket handles.

### DIFF
--- a/components/fs/nfs/posix.c
+++ b/components/fs/nfs/posix.c
@@ -314,7 +314,6 @@ long sys_socket(va_list ap)
 	return -1;
     }
 
-
     int socket_handle = tcp_socket_create();
     if (socket_handle != -1) {
         socket_refcount[socket_handle]++;
@@ -354,7 +353,7 @@ long sys_close(va_list ap)
     long fd = va_arg(ap, int);
 
     assert(fd_active[fd]);
-    
+
     int socket_handle = fd_socket[fd];
 
     assert(socket_handle >= 0);
@@ -380,7 +379,6 @@ long sys_dup3(va_list ap)
     (void)flags;
 
     assert(fd_active[oldfd]);
-    
     int oldfd_socket_handle = fd_socket[oldfd];
 
     assert(oldfd_socket_handle >= 0);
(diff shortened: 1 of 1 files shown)


components/fs/nfs/tcp.c
```diff
@@ -357,11 +359,12 @@ err_t socket_connected(void *arg, struct tcp_pcb *tpcb, err_t err)
     assert(socket->state == socket_state_connecting);
 
     socket->state = socket_state_connected;
-    
+
     tcp_sent(tpcb, socket_sent_callback);
     tcp_recv(tpcb, socket_recv_callback);
 
     tpcb->so_options |= SOF_KEEPALIVE;
+    socket->sock_tpcb->keep_intvl = 1000;
```
Ivan-Velickovic, line 367: Can we have something like `#define LWIP_KEEPALIVE_INTERVAL 1000` in `lwipopts.sh`?

Ivan-Velickovic: And also a comment next to the define that says it's 10s etc.

components/fs/nfs/lwip_include/lwipopts.h
```diff
@@ -21,7 +21,9 @@
 #define LWIP_ICMP 1
 #define LWIP_RAND rand
 #define LWIP_DHCP 1
-
+#define LWIP_TCP_KEEPALIVE  1
+#define TCP_KEEPIDLE_DEFAULT    1000
```
Ivan-Velickovic, line 25: Should this be called `TCP_KEEPINTVL_DEFAULT` instead?

wom-bat:
> I'm not sure we want to change these so dramatically without proper justification. It might incidentally fix an issue but not fix the real cause. Do you know anything about appropriate times for keepalive @wom-bat?

Standard says between 10 minutes and 2 hours.  I'd suggest around 15 minutes for our use cases.
See https://webhostinggeeks.com/howto/tcp-keepalive-recommended-settings-and-best-practices/

Ivan-Velickovic:
After talking to @JE-Archer we'll just merge the patch that adds the error check. @JE-Archer believes there is a deeper issue and playing with the parameters of lwIP is just hiding that, so he's looking into that.

Ivan-Velickovic (approved):

PR #101, opened 2024-08-21
Initialise submodules on demand

wom-bat:
Add dependencies so that Make knows that submodules need to be downloaded, and add rules to download them.

### DIFF
--- a/examples/vmm/vmm-dev.mk
+++ b/examples/vmm/vmm-dev.mk
@@ -129,10 +129,6 @@ clobber:: clean
 # How to build libvmm.a
 include ${LIBVMM}/vmm.mk
 
-# Add an extra dependency for libvmm.a
-# Note: this should be in ${LIBVMM}/vmm.mk
-libvmm/virtio/block.o: ${SDDF}/include
-
 ${SDDF}/include ${LIBVMM}/vmm.mk:
 	cd ${LIONSOS}; git submodule update --init $(dir $@)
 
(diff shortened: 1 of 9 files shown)


examples/vmm/vmm-dev.mk
```diff
@@ -124,3 +128,11 @@ clobber:: clean
 
 # How to build libvmm.a
 include ${LIBVMM}/vmm.mk
+
+# Add an extra dependency for libvmm.a
+# Note: this should be in ${LIBVMM}/vmm.mk
+libvmm/virtio/block.o: ${SDDF}/include
```
Ivan-Velickovic, line 134: I noticed when working on libvmm that updating the headers did not cause objects to be rebuilt, I think we should add this as well as `libvmm/include` to `libvmm/vmm.mk` in another PR.

wom-bat: libvmm/include should not be necessary providing -MD is in CFLAGS.

wom-bat: I've just looked at vmm.mk --- it has a different bug, stemming from when object files were put in a libvmm directory, not a src directory -- the result is the dependency files that track headers were never included.
PR for libvmm coming soon.

Ivan-Velickovic: Still seems weird that we have this line. Do we need it anymore after the fix to libvmm?

Ivan-Velickovic:
@JE-Archer ran into an issue where he manually updated a submodule dependency (sDDF) and running `make` kept resetting it to the old commit.

Another problem I've just realised is the scenario where someone's got an old checkout of LionsOS, goes to an example and runs `git pull` and then `make`. None of the submodules will be updated since `git pull` doesn't do that but there is also no `make submodules` command they can run to update the submodules.

It might be worth going back to having an explicit `make submodules` command.

PR #103, opened 2024-08-23
FAT (File Allocation Table) file system for Lions OS

Cheng-Li1:
The FAT (File Allocation Table) file system for Lions OS is an asynchronous file system built upon Elm-Chan's FatFs - Generic FAT Filesystem Module ([FatFs website](http://elm-chan.org/fsw/ff/)). The fatfs for Lions OS supports the FAT12, FAT16, FAT32, and exFAT file system formats.

This implementation also incorporates changes from the pull request titled "Make FS protocol more explicit." The communication protocol between fatfs and clients is based on this enhanced FS protocol.

The transition from a synchronous file system to an asynchronous one is achieved using coroutine. The file system operates as follows: The main coroutine is responsible for receiving and delegating requests and responses to worker coroutines, which in turn invoke the actual file system functionalities, with yielding to other coroutines when encountering blocking operation.


Ivan-Velickovic:
It should be noted that this PR also depends on at least two changes that aren't in sDDF:
* the change of `SDDF_TRANSFER_SIZE` to 512. I think it should be kept to 4096.
* the removal of the virtualiser check that the offset passed by the client is on a transfer size boundary. That change does make sense to me.

Ivan-Velickovic:
It will take multiple rounds of review to get this merged but some low-hanging fruit you can work towards is:
* using libmicrokitco
* having consistent style with all the other code, e.g `Client_CH` should be `CLIENT_CH`, the includes should not use relative paths.
* documenting the design of the FAT file system in the LionsOS design doc.
* Where does `ff15` come from? That needs to be documented somewhere but I'm not sure what the best place for it is.
* Adding a basic Python script we can use to test FATFS and file systems in general.
* Could you explain the relationship between, `fatfs_op.c`, `fatfs_event.c` and `fs_diskio.c`?

Ivan-Velickovic:
One thing I'm a bit confused about is how MicroPython knows whether to use FATFS or NFS, right now in `micropython.c` it still does `init_nfs` but offline we were using FATFS just fine. I don't see anything in the Makefile to tell MicroPython to use FATFS instead of NFS? From memory MicroPython only supports one VFS but I might be wrong.

Cheng-Li1:
> One thing I'm a bit confused about is how MicroPython knows whether to use FATFS or NFS, right now in `micropython.c` it still does `init_nfs` but offline we were using FATFS just fine. I don't see anything in the Makefile to tell MicroPython to use FATFS instead of NFS? From memory MicroPython only supports one VFS but I might be wrong.

Fatfs and nfs follow the same set of protocol so micropython does not need to who it is talking to.

Ivan-Velickovic:
> Where does ff15 come from? That needs to be documented somewhere but I'm not sure what the best place for it is.

My bad it's in the PR description.

Cheng-Li1:
> It will take multiple rounds of review to get this merged but some low-hanging fruit you can work towards is:
> 
> * using libmicrokitco
> * having consistent style with all the other code, e.g `Client_CH` should be `CLIENT_CH`, the includes should not use relative paths.
> * documenting the design of the FAT file system in the LionsOS design doc.
> * Where does `ff15` come from? That needs to be documented somewhere but I'm not sure what the best place for it is.
> * Adding a basic Python script we can use to test FATFS and file systems in general.
> * Could you explain the relationship between, `fatfs_op.c`, `fatfs_event.c` and `fs_diskio.c`?

Good suggestions, I will be working on those. One minor issue is libmicrokitco seems to still have some bugs and I would switch to use libmicrokitco once it works correctly.

components/fs/fat/fatfs_op.c
```diff
@@ -0,0 +1,798 @@
...
+    uint32_t i;
+    for (i = 0; i < MAX_OPENED_FILENUM; i++) {
+        if (file_status[i] == FREE) {
+            return i;
+        }
+    }
+    return i;
+}
+
+uint32_t Find_FreeDir(void) {
```
Ivan-Velickovic, line 130: Another style thing, all function/variable names should be snake_case.

components/micropython/micropython.c
```diff
@@ -92,8 +92,11 @@ void t_mp_entrypoint(void) {
     gc_init(heap, heap + sizeof(heap));
     mp_init();
 
-    init_nfs();
     init_networking();
+    // initialisation of the filesystem utilises the event loop and the event
+    // loop unconditionally tries to process incoming network buffers; therefore
+    // the networking needs to be initialised before initialising the fs
+    init_nfs();
```
Ivan-Velickovic, line 99: I think this should be probably be called something like `mp_fs_init`.

Cheng-Li1:
> * the change of `SDDF_TRANSFER_SIZE` to 512. I think it should be kept to 4096.

Let me add a bit more context to avoid confusion. FATFS is flexible and can work with various sector sizes (e.g., 512 bytes, 1024 bytes, 2048 bytes, 4096 bytes, etc.). However, the sector size that the FAT file system operates on is determined at the time of formatting. When the FAT file system is mounted, the file system driver reads the metadata from the first sector (sector 0), which includes the sector size.

The problem here is that the storage medium has a physical sector size of 512 bytes, and the FAT file system on it is formatted with a 512-byte sector size in mind. The block virtualizer, however, tries to read multiple sectors at a time and emulates a 4096-byte sector size to improve performance. This approach can confuse the file system, as it expects to read one sector and provides a 512-byte buffer, but 4096 bytes are written to that buffer, which inevitably leads to incorrect results.

However, since the file system is aware that the block virtualizer is emulating a 4096-byte sector, it is possible to construct a translation layer to handle this correctly. For example, a read of sector 8 in 512-byte sectors could be mapped to a read of sector 1 by the block virtualizer, followed by copying the first 512 bytes of data. Similarly, a write to a single sector could be handled using a read-modify-write cycle.

Given that the file system currently has only minimal caching mechanisms and reads/writes one sector at a time due to the lack of dynamic memory allocation, I cannot think of any better way to address this issue than using an inefficient translation layer as described.

However, I am open to suggestions if there is a more efficient approach to handle this situation. 

Btw, to get the fileio example working, simply creating a disk with a 4096-byte sector size should also work, as long as the sector size is the same as BLK_TRANSFER_SIZE defined in blk/queue.h in SDDF.

Cheng-Li1:
> * Could you explain the relationship between, `fatfs_op.c`, `fatfs_event.c` and `fs_diskio.c`?
fatfs_event.c defines the init process, event handling and is responsible for init, assign request, and wake up worker coroutines. 
fatfs_op.c defines the wrapper functions that the worker coroutine is executing, those wrapper functions call the file system operations declared in ff15/source/ff.h. The wrapper functions extract function variables from the args that the event coroutine passes to the worker coroutine, and run the variables with a bunch of sanity checks, then finally pass the variable to file system operations defined in ff15/source/ff.h. The fatfs_event.c and fatfs_op.c's interaction is via the operation_functions function array, fatfs_op.c define those functions and fatfs_event.c assign those functions(requests) to worker coroutines.
fs_diskio.c contains the disk functions that the file system operations call. For example, if the fs want to read a certain sector on the disk, the fs would call disk_read. The disk operation is implemented as enqueuing a disk request to the sddf queue between fs and blk virt and block the worker coroutine. When the fat_event.c receive a response from the blk virt, it wakes up the worker coroutine that blocks on that request.
co_helper.c is an abstraction layer to abstract away the differences of two coroutine libraries that I am using and provide a constant set of coroutine functions to the file system.

examples/fileio/board/qemu_virt_aarch64/fileio.system
```diff
@@ -0,0 +1,295 @@
...
+        <map mr="blk_client_data" vaddr="0x40_800_000" perms="rw" cached="true" setvar_vaddr="fs_metadata" />
+
+        <map mr="nfs_command_queue" vaddr="0x7_800_000" perms="rw" cached="true" setvar_vaddr="fatfs_command_queue" />
+        <map mr="nfs_completion_queue" vaddr="0x7_810_000" perms="rw" cached="true" setvar_vaddr="fatfs_completion_queue" />
+        <map mr="shared_fs_micropython" vaddr="0x7_900_000" perms="rw" cached="true" setvar_vaddr="client_data_addr"/>
+
+        <map mr="coroutine_stack1" vaddr="0xA0_000_000" perms="rw" cached="true" setvar_vaddr="coroutine_stack_one" />
+        <map mr="coroutine_stack2" vaddr="0xB0_000_000" perms="rw" cached="true" setvar_vaddr="coroutine_stack_two" />
+        <map mr="coroutine_stack3" vaddr="0xC0_000_000" perms="rw" cached="true" setvar_vaddr="coroutine_stack_three" />
+        <map mr="coroutine_stack4" vaddr="0xD0_000_000" perms="rw" cached="true" setvar_vaddr="coroutine_stack_four" />
```
Ivan-Velickovic, line 228: This should be global variables instead of memory regions in my opinion. The stacks should be small enough that having it as global data is fine.

Ivan-Velickovic:
Another thing you can work on while we review is having the web server example use the FAT file system instead of NFS. This would be a good test of the FAT FS. The web server also runs on QEMU.

examples/fileio/board/qemu_virt_aarch64/fileio.system
Cheng-Li1: But I think by doing this way we would have guard pages to detect stack overflow from the coroutine stacks?

Cheng-Li1:
> Another thing you can work on while we review is having the web server example use the FAT file system instead of NFS. This would be a good test of the FAT FS. The web server also runs on QEMU.

Right, I have put this on my to-do list.

Ivan-Velickovic:
I see that FATFS depends on libc, what functionality from libc does it depend on?

Ivan-Velickovic:
The MicroPython port, understandably, mentions NFS in various places where it shouldn't. I think the only two places are in:
* `micropython/fs_helpers.c` for the data/command/completion regions.
* `micropython/micropython.c` for the MicroPython VFS initialisation.

Cheng-Li1:
> I see that FATFS depends on libc, what functionality from libc does it depend on?

Memory functions like memcpy, memset. Str functions like strlen. math functions like isnan and isinf. The header files included are:
string.h, math.h, stdint.h, stdbool.h, stdarg.h.

Ivan-Velickovic:
> It should be noted that this PR also depends on at least two changes that aren't in sDDF yet:
> 
> * the change of `SDDF_TRANSFER_SIZE` to 512. I think it should be kept to 4096.
> * the removal of the virtualiser check that the offset passed by the client is on a transfer size boundary. That change does make sense to me.

As discussed offline, both changes aren't necessary/wanted and so we don't need to make any changes to sDDF for FATFS.

Cheng-Li1:
> > It should be noted that this PR also depends on at least two changes that aren't in sDDF yet:
> > 
> > * the change of `SDDF_TRANSFER_SIZE` to 512. I think it should be kept to 4096.
> > * the removal of the virtualiser check that the offset passed by the client is on a transfer size boundary. That change does make sense to me.
> 
> As discussed offline, both changes aren't necessary/wanted and so we don't need to make any changes to sDDF for FATFS.

Those two changes are no longer required as the translation layer is implemented and tested. Now no changes are needed to be done for sddf to run the fileio example.

Ivan-Velickovic:
Now that the bugs regarding libmicrokitco are resolved, can you remove all the other co-routine stuff and just leave libmicroktico?

examples/fileio/fileio.mk
```diff
@@ -0,0 +1,179 @@
...
+		   LD=$(LD) \
+		   CPU=$(CPU) \
+		   BUILD_DIR=$(abspath .) \
+		   CONFIG=$(MICROKIT_CONFIG) \
+	       MICROKIT_SDK=$(MICROKIT_SDK) \
+		   MICROKIT_BOARD=$(MICROKIT_BOARD) \
+		   LIBC_DIR=$(abspath $(BUILD_DIR)/musllibc) \
+		   BUILD_DIR=$(abspath .) \
+		   TARGET=$(TARGET) \
+		   USE_LIBMICROKITCO=1
```
Ivan-Velickovic, line 136: E.g this should not be an option

components/fs/fat/co_helper.c
Ivan-Velickovic, line 1: Same with these files.

Cheng-Li1: Yes, those are changes I am planning to make tomorrow, along with some refinement for the recent implemented translation layer.

Ivan-Velickovic:
@Cheng-Li1 James' FS protocol changes have been merged so you might want to rebase on main.

Cheng-Li1:
> @Cheng-Li1 James' FS protocol changes have been merged so you might want to rebase on main.

Done, hope it is the last time I do force push.

components/fs/fat/fs_diskio.c
```diff
@@ -0,0 +1,290 @@
...
+			int32_t aligned_sector = count - unaligned_head_sector - unaligned_tail_sector;
+			sddf_count += DIV_POWER_OF_2(aligned_sector, sector_per_transfer);
+			
+			assert(MUL_POWER_OF_2(sddf_count, BLK_TRANSFER_SIZE) <= MAX_CLUSTER_SIZE);
+
+			#ifdef FS_DEBUG_PRINT
+			sddf_printf("blk_enqueue_read pre adjust: addr: 0x%lx sector: %u, count: %u ID: %d\n", read_data_offset, sector, count, handle);
+			sddf_printf("blk_enqueue_read after adjust: addr: 0x%lx sector: %u, count: %d ID: %d\n", read_data_offset, sddf_sector, sddf_count, handle);
+			#endif
+			blk_enqueue_req(blk_queue_handle, BLK_REQ_READ, read_data_offset, sddf_sector, sddf_count, handle);
```
Ivan-Velickovic, line 172:
```suggestion
			int err = blk_enqueue_req(blk_queue_handle, BLK_REQ_READ, read_data_offset, sddf_sector, sddf_count, handle);
			assert(!err);
```

To confirm the invariant that the coroutine only ever enqueues a single block request.

components/fs/fat/fs_diskio.c
```diff
@@ -0,0 +1,290 @@
...
+
+			blk_request_pushed = true;
+			co_block();
+
+			res = (DRESULT)(uintptr_t)co_get_args();
+			memcpy(buff, (void*)(read_data_offset + blk_data_region + sector_size * MOD_POWER_OF_2(sector, sector_per_transfer)), sector_size * count);
+			#ifdef FS_DEBUG_PRINT
```
Ivan-Velickovic, line 179: The `#ifdefs` should always be like this:

```suggestion
+#ifdef FS_DEBUG_PRINT
```

they should always be aligned to the start. You can see examples of this in sDDF and LionsOS.

components/fs/fat/FATFS_README.md
```diff
@@ -9,51 +9,48 @@ ASYNCFATFS provides and relies on a fully asynchronous file system API and block
 To initiate a request to the file system, the client must enqueue one or more requests using the interfaces defined in fs_protocol.h, and notify the file system accordingly. The asynchronous nature ensures that the client does not need to block while waiting for the operation to complete.
 
 # How FATFS Is Modified to Be Asynchronous
-The asynchronous design of ASYNCFATFS is very similar to Linux’s io_uring, which efficiently handles non-blocking I/O operations using an event-driven model. The system follows a coroutine-based event-worker design, where requests are processed without blocking, and I/O operations are handled asynchronously.
+The asynchronous design of ASYNCFATFS is very similar to Linux’s io_uring, which efficiently handles non-blocking I/O operations using an event-driven model. The system follows a thread-based event-worker design, where requests are processed without blocking, and I/O operations are handled asynchronously.
```
Ivan-Velickovic, line 12: Why change this to thread-based? Wouldn't coroutine-based be more accurate?

Ivan-Velickovic: I just looked at the commit message which answers my question. Although we need to be clear somewhere that these are not kernel-threads. Perhaps at the top of the README it can say that when we talk about 'threads' we are talking about co-operative multithreading.

components/fs/fat/fatfs_config.h
```diff
@@ -0,0 +1,34 @@
...
+ *  This def control whether the memory address passed to the blk device driver should be strict aligned to the 
+ *  BLK_TRANSFER_SIZE and if the BLK_TRANSFER_SIZE should be set to 4KB even if the file system is formatted with
+ *  a different sector size. This setting is for future benchmarking purpose. When using with unmodified SDDF, this should
+ *  always be enabled even though it may cause some performance degradation.
+ *  To enable benchmarking for zero copy, low overhead write, aside from delete the def below, the .system file and blk_config.h
+ *  must be modified to correctly map shared memory between micropython and file system and fs_metadata to blk virt and blk driver.
+ *  SDDF need to be modified as well to remove page alignment check and make sure BLK_TRANSFER_SIZE is equal to the blk device sector
+ *  size.
+ */
+#define MEMBUF_STRICT_ALIGN_TO_BLK_TRANSFER_SIZE
```
Ivan-Velickovic, line 34: I'm not sure if we want this to be a configuration option.

Ivan-Velickovic: sDDF sets the policy for block transfer size right now so I don't think the file system should be encouraging people to change it.

Cheng-Li1: That is only for comparing the overhead of strict alignment in future benchmarking. I think I have left a comment above that option.

Ivan-Velickovic: I'm not sure if we should be prematurely generalising for benchmarking that hasn't happened yet.

components/fs/fat/fatfs_config.h
Ivan-Velickovic, line 1: All the configuration options should begin with `FATFS_`, e.g `DATA_REGION_SIZE` is confusing since it's not clear whether it's to do with sDDF or not.

components/fs/fat/fatfs_config.h
Cheng-Li1: Right, I will change it later today, I am trying to get sdmmc driver compiled with sddf_blk and testing it with the file system first, and then I will fix the issues you have mentioned.

Ivan-Velickovic:
Couple notes when I compiled it for something I'm working on locally:
* trailing whitespace needs to be removed
* `blk_config.h` should always be included as `#include <blk_config.h>` in case the user does not have a `config` directory
* `#include <fs/protocol.h>` -> `#include <lions/fs/protocol.h>`

Ivan-Velickovic:
There's also some unused variables:
```
/Users/ivanv/ts/lionsos_tutorial/lionsos/components/fs/fat/io.c:49:13: error: unused variable 'stat'
    DSTATUS stat;
            ^
/Users/ivanv/ts/lionsos_tutorial/lionsos/components/fs/fat/io.c:50:9: error: unused variable 'result'
    int result;
        ^
/Users/ivanv/ts/lionsos_tutorial/lionsos/components/fs/fat/io.c:94:13: error: unused variable 'stat'
    DSTATUS stat;
            ^
/Users/ivanv/ts/lionsos_tutorial/lionsos/components/fs/fat/io.c:95:9: error: unused variable 'result'
    int result;
        ^
```

These should be caught with compiler flags ideally. We should add `-Wall -Werror` to the Makefile.

components/fs/fat/Makefile
```diff
@@ -0,0 +1,107 @@
...
+	printf.o \
+	putchar_debug.o \
+	assert.o
+
+LIBMICROKITCO_OBJ := libmicrokitco/libmicrokitco.a
+LIBMICROKITCO_PATH := $(LIONSOS)/dep/libmicrokitco
+LIBMICROKITCO_OPT_PATH := $(LIONSOS)/components/fs/fat
+
+LIBMICROKITCO_OBJS := $(FATFS_OBJECT_DIR)/$(LIBMICROKITCO_OBJ)
+CPPFLAGS := -DUSE_LIBMICROKITCO
```
Ivan-Velickovic, line 65: Not necessary anymore.

components/fs/fat/event.c
```diff
@@ -0,0 +1,281 @@
...
+
+// The notified function requires careful management of the state of the file system
+/*
+  The filesystems should be blockwait for new message if and only if all of working
+  threads are either free(no tasks assigned to them, no pending replies) or blocked in diskio.
+  If the filesystem is blocked here and any working threads are free, then the fatfs_command_queue 
+  must also be empty.
+*/
+void notified(microkit_channel ch) {
+    LOG_FATFS("FS IRQ received::%d\n", ch);
```
Ivan-Velickovic, line 163:
```suggestion
    LOG_FATFS("FS IRQ received::%d\n", ch);
```

components/fs/fat/event.c
```diff
@@ -0,0 +1,281 @@
...
+
+    uint64_t command_queue_size;
+    uint64_t completion_queue_size;
+
+    uint32_t fs_request_dequeued = 0;
+    uint32_t fs_response_enqueued = 0;
+
+    /**
+      I assume this big while loop is the confusing and critical part for dispatching threads and send back the results.
+    **/
```
Ivan-Velickovic, line 182: This comment confuses me.

components/fs/fat/event.c
```diff
@@ -0,0 +1,281 @@
...
+        {
+            blk_resp_status_t status;
+            uint16_t success_count;
+            uint32_t id;
+            // Get current element numbers in blk response queue
+            uint32_t len = blk_queue_length_resp(blk_queue_handle);
+
+            while (len > 0) {
+                // This id is the index to the request pool
+                blk_dequeue_resp(blk_queue_handle, &status, &success_count, &id);
```
Ivan-Velickovic, line 193: Always check the return value of enqueuing and dequeuing with an assert at least. If it is known to be correct then just an assert is fine.

Having an assert catches a surprising amount of errors for misconfigured setups.

Ivan-Velickovic: There's only one occurrence of `blk_dequeue_resp` that I can find but 8 `blk_enqueue_req` that needs to be checked as well.

examples/fileio/src/config/blk_config.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include <sddf/util/string.h>
+#include <sddf/blk/queue.h>
+#include <sddf/blk/storage_info.h>
+
+#define BLK_NUM_CLIENTS 1
+
+#define BLK_NAME_CLI0                      "fatfs"
+
+#define BLK_QUEUE_SIZE_CLI_FATFS            16
```
Ivan-Velickovic, line 16: Why only 16 buffers? I assume because each worker cothread only enqueues one request at a time?

components/fs/fat/op.c
```diff
@@ -0,0 +1,768 @@
...
+    }
+
+    args->status = (RET == FR_OK) ? FS_STATUS_SUCCESS : FS_STATUS_ERROR;
+}
+
+// Mode attribute
+#define Directory 0040000
+#define Regularfile 0100000
+#define Blockdevice 0060000
+#define Socket 0140000
```
Ivan-Velickovic, line 375: Capital letters, but also I don't see uses of any of these.

components/fs/fat/io.c
```diff
@@ -0,0 +1,241 @@
...
+        LOG_FATFS("blk_enqueue_syncreq\n");
+        blk_enqueue_req(blk_queue_handle, BLK_REQ_FLUSH, 0, 0, 0, microkit_cothread_my_handle());
+        blk_request_pushed = true;
+        block();
+        res = (DRESULT)(uintptr_t)microkit_cothread_my_arg();
+    }
+    return res;
+}
+
+#ifdef MEMBUF_STRICT_ALIGN_TO_BLK_TRANSFER_SIZE
```
Ivan-Velickovic, line 117: Do we still want this?

components/fs/fat/config.h
```diff
@@ -0,0 +1,39 @@
...
+
+// Maximum opened directories
+#define MAX_OPENED_DIRNUM 16
+
+// The number of worker thread, if changes are needed, initialisation of the thread pool must be changes as well
+#define WORKER_THREAD_NUM 4
+
+#define THREAD_NUM (WORKER_THREAD_NUM + 1)
+
+#define WORKER_THREAD_STACKSIZE 0x40000
```
Ivan-Velickovic, line 27: These need `FATFS_*` prefix

Ivan-Velickovic:
Ignore the trailing whitespace check for everything in `dep/ff15` but there's lots in `components/fs/fat`.

Ivan-Velickovic (approved):
Follow up work in https://github.com/au-ts/lionsos/issues/127.

Thanks to @Cheng-Li1 for implementing it, and @JE-Archer for getting it over the finish line.

PR #105, opened 2024-08-27
EXEC_MODULE for micropython needs module name, not path

wom-bat:
Without this change, the micropython server cannot find the webserver module, as its name is given with respect to the build environment, not the running environment.


Ivan-Velickovic:
Tested locally and it works now, now that QEMU support is available for both the webserver and Kitty, we should add runtime tests to catch regressions.

PR #106, opened 2024-08-28
Async repl

wom-bat:
This is a straw-man proof of concept.  It allows interacting with a REPL for examples/webserver on the qemu_virt_aarch64 platform 

I can create a file 'extra.py' on the NFS server, that contains:
```
    app.url_map = [(['GET'], p, action)] + app.url_map

counter = 0

def foo(request):
    global counter
    head = '''
    <html>
      <body>
        <h1>Foo</h1>
    '''
    tail = '''
      </body>
    </html>
'''
    counter = counter + 1
    body = head + str(counter) + tail
    headers = {'Content-Type': 'text/html'}
    return Response(body=body, headers = headers)
```

Then on the commandline do:
```
--> import sys
--> sys.path.append('micropython')
--> from extra import foo, add_action
--> add_action(app, '/foo', foo)
```

Then visiting the webserver's `/foo` prints out the counter value.

It'd be trivial to extend to deleting the path and modifying existing ones.

The Asyncio repl is a bit limited compared with the standard one: the biggest limitation is one can use only single line commands.

If this is considered a good idea, I'll add it to the other webserver example platforms.


Ivan-Velickovic:
Seems like a good idea to me, given that it's completely optional.

The main limitation I see is that it needs the script that you are running to allow use aiorepl, but I guess that's a limitation with MicroPython itself and not this PR.

I'll play around with the PR and see how well it works.

Ivan-Velickovic:
One question I have is, is there anyway to stop a script that's running if you're in the async REPL?

wom-bat:
> One question I have is, is there anyway to stop a script that's running if you're in the async REPL?
Yes there is.

From the async repl, it feels (almost) exactly the same as a standard repl, except that other things are running in the background.

You can either set it up to share the initial context (as I did) or to have its own.  If it has the initial context, then you can do:
```
app.shutdown()
```
to stop the webserver.

If you start new coroutines from the repl, you can use all asyncio's methods to control them

Ivan-Velickovic:
Okay, is there any reason why this is a draft?

Ivan-Velickovic:
We need to start documenting what sub-set of MicroPython features we support. This being one of them.

PR #108, opened 2024-08-29
Add more responses to kitty UI

MichaelM-UNSW:
*NEEDS TO BE TESTED BEFORE MERGE*

The purpose of this PR, was to modify the current kitty client, running on LionsOS, to be able to interact with the (proper) existing kitty server. This involved adapting the existing client UI to be able to respond to ALL possible responses from the kitty server.

Changes:
 - Modified client to strip response messages of their code and print the message at the "center" of the screen
 - Added extra "font" instance (font35.py) to print the server time in smaller text (I chose this as I believe it looks slightly better than the original font size) 
 - Modified makefiles + client code to run immediately on startup as well as to point at the fridge host address
 - Modified testing server (server.py) to repeatedly cycle through all the possible responses that the fridge server could respond with (cycles through 1 message with every tap)


components/fs/nfs/lwip_include/lwipopts.h
Kswin01, line 1: Maybe drop the commits that were merged in from https://github.com/au-ts/lionsos/pull/100 just for this PR.

examples/kitty/board/odroidc4/kitty.system
```diff
@@ -329,7 +329,7 @@
 
     <channel>
         <end pd="i2c_driver" id="0" />
-        <end pd="i2c_virt" id="1" />
+        <end pd="i2c_virt" id="2" />
```
Kswin01, line 332: What was the need for this change? Was it something to do with the sddf i2c changes?

examples/kitty/board/odroidc4/config.py
```diff
@@ -1,4 +1,5 @@
 display_width = 1920
 display_height = 1080
+TRUE_DISPLAY_WIDTH = 1024
```
Kswin01, line 3: Can we use the same value for the qemu config?

examples/kitty/client/kitty.py
```diff
@@ -30,14 +32,20 @@
 writer_stream = None
 stdin_ready = True
 
-IP_ADDRESS = ""
-PORT = 3738
+CURR_NUM_TAPS = 0
```
Kswin01, line 35: Remove these variables if we decide not to track current taps.

examples/kitty/client/kitty.py
```diff
@@ -253,21 +265,44 @@ async def read_from_server():
...
+            else:
+                message_display_time = 3
+
+            # Printing "-$1" if there has been a purchase
+            if words[0] == '200':
+                display.rect(0, 500, 2000, 2000, 0x0, True)
+                wri50.set_textpos(display, 500, (TRUE_DISPLAY_WIDTH - wri50.stringlen(f'-$1'))//2)
+                
+                wri50.printstring(f"-$1")
+                display.show()
```
Kswin01, line 302: Can we have one `display.show()` at the end of this conditional to make sure the framebuffer is updated once.

examples/kitty/client/kitty.py
```diff
@@ -325,23 +361,28 @@ async def main():
...
+    wri35 = CWriter(display, font35)
+    HOST = host
     asyncio.run(main())
 
 
-help_prompt()
+# help_prompt() 
+
+# Comment when in dev
+run()
```
Kswin01, line 388: Not sure if we'd want to keep this here for this PR.

examples/kitty/client/kitty.py
```diff
@@ -325,23 +361,28 @@ async def main():
...
 
 def help_prompt():
-    print("\nWelcome to Kitty!\nUsage: kitty.run(String IP_ADDRESS).\n\
-            IP_ADDRESS: The IP address of the kitty server.")
+    print("\nWelcome to Kitty!\nUsage: kitty.run(String HOST).\n\
+            HOST: The host name / IP address of the kitty server.")
 
 
-def run(ip="172.16.0.2"):
+def run(host="fridge.keg.cse.unsw.edu.au"):
```
Kswin01, line 368: And not too sure if we want to merge our host name into main, although that is what we've done in the past.

components/fs/nfs/lwip_include/lwipopts.h
```diff
@@ -21,7 +21,6 @@
 #define LWIP_ICMP 1
 #define LWIP_RAND rand
 #define LWIP_DHCP 1
-
```
Kswin01, line 24: Minor whitespace

examples/kitty/board/odroidc4/kitty.system
Kswin01, line 332: Also reflect this change in `board/qemu` `kitty.system` file

examples/kitty/board/odroidc4/config.py
Kswin01, line 3: Add this variable in `board/qemu` `config.py` file

examples/kitty/client/kitty.py
```diff
@@ -30,14 +32,18 @@
 writer_stream = None
 stdin_ready = True
 
-IP_ADDRESS = ""
-PORT = 3738
+CURR_USERNAME = ""
```
Kswin01, line 35: Can we remove these variables completely if they're unused?

examples/kitty/client/kitty.py
```diff
@@ -325,23 +359,25 @@ async def main():
...
-    wri = CWriter(display, font)
-    IP_ADDRESS = ip
+    wri50 = CWriter(display, font50)
+    wri35 = CWriter(display, font35)
+    HOST = host
     asyncio.run(main())
 
 
-help_prompt()
+run()
```
Kswin01, line 383: Can we change this back to `help_prompt()`?

PR #111, opened 2024-09-06
Use lib sDDF LWIP in Micropython and NFS

Courtney3141:
This PR uses the sDDF LWIP library to interface between sDDF and LWIP in both micropython and NFS components. This significantly reduces the repeated code in both these components and the echo server. As explained in the corresponding sDDF PR: https://github.com/au-ts/sddf/pull/238 there are still two remaining issues with building the library using a makefile snippet, so in this PR the library is built in each components makefile without the use of the snippet.

Both examples successfully build and run, and the performance has been verified to not be degraded. 

Depends on https://github.com/au-ts/sddf/pull/238.


components/fs/nfs/tcp.c
```diff
@@ -71,222 +52,39 @@ typedef struct {
...
+    net_queue_init(&tx_handle, net_config.tx.free_queue.vaddr, net_config.tx.active_queue.vaddr, net_config.tx.num_buffers);
+    net_buffers_init(&tx_handle, 0);
 
-    if (notify_tx && net_require_signal_active(&state.tx_queue)) {
-        net_cancel_signal_active(&state.tx_queue);
-        notify_tx = false;
-        if (!microkit_have_signal) microkit_deferred_notify(net_config.tx.id);
-        else if (microkit_signal_cap != BASE_OUTPUT_NOTIFICATION_CAP + net_config.tx.id) microkit_notify(net_config.tx.id);
-    }
+    sddf_lwip_init(&lib_sddf_lwip_config, &net_config, &timer_config, rx_handle, tx_handle, NULL, netif_status_callback, NULL);
```
Courtney3141, line 85: Might make sense here to pass a pointer to NFS's output function (`printf`?) as the `lib_sddf_lwip`'s `err_output` function.

components/micropython/micropython.c
```diff
@@ -86,8 +99,15 @@ void t_mp_entrypoint(void) {
     mp_init();
 
     if (net_enabled) {
-        init_networking();
+        net_queue_init(&net_rx_handle, net_config.rx.free_queue.vaddr, net_config.rx.active_queue.vaddr, net_config.rx.num_buffers);
+        net_queue_init(&net_tx_handle, net_config.tx.free_queue.vaddr, net_config.tx.active_queue.vaddr, net_config.tx.num_buffers);
+        net_buffers_init(&net_tx_handle, 0);
+
+        sddf_lwip_init(&lib_sddf_lwip_config, &net_config, &timer_config, net_rx_handle, net_tx_handle, NULL, netif_status_callback, NULL);
```
Courtney3141, line 106: Similarly here too, we could pass Micropython's output function. Also it may make sense here to use the default `lib_sddf_lwip`'s `netif_callback` function as it basically does the same thing (currently prints out the obtained IP and address of the client, however we could change it to print the client's `microkit_name` too).

Courtney3141 (commented):
Other than my suggestions, everything seems correct to me.

Courtney3141:
Tested the changes on the webserver. On release mode the command:
`time wget --recursive --no-clobber --page-requisites --html-extension --convert-links --restrict-file-names=windows --domains $WEBSERVER_IP --no-parent $WEBSERVER_IP`
which downloads all pages took 1:50.21 on current main and 1:45.35 after these changes.

PR #116, opened 2024-09-18
NFS: Fix TCP socket leak

JE-Archer:


components/fs/nfs/tcp.c
```diff
@@ -298,6 +306,7 @@ static err_t socket_recv_callback(void *arg, struct tcp_pcb *tpcb, struct pbuf *
     socket_t *socket = arg;
     assert(socket != NULL);
     int socket_index = (int)(socket - sockets);
```
Ivan-Velickovic, line 308: Use `socket_id` on this line?

Ivan-Velickovic:
```suggestion
    int socket_index =  socket_id(socket);
```

PR #119, opened 2024-09-26
webserver: try harder to resolve URLs

JE-Archer:
Currently the webserver system resolves a URL `path/to/name` to either exactly `path/to/name` or to `path/to/name/index.html` if `name` is a directory. This PR changes the system to also try to resolve `path/to/name` to `path/to/name.html`, `path/to/name.htm` and `path/to/name.xhtml`, in line with standard practice. The PR also fixes a small issue with checking for malicious directory traversal, where it would reject valid filenames that included `..`.

### DIFF
--- a/examples/webserver/webserver.py
+++ b/examples/webserver/webserver.py
@@ -71,6 +71,67 @@ def format_http_date(timestamp):
     return formatted_date
 
 
+async def resolve(relative_path):
+    # Standard html file extensions to try suffixing to the requested
+    # file if the file does not exist.
+    html_extensions = ['.html', '.htm', '.xhtml']
+
+    def is_dir(stat):
+        return stat[0] & 0o170000 == 0o40000
+
+    async def try_stat(path):
+        try:
+            return await fs_async.stat(path)
+        except:
+            return None
+
+    async def try_suffices(path, suffices):
+        # TODO: maybe stat these concurrently
+        for suffix in suffices:
+            suffixed_path = path + suffix
+            stat = await try_stat(suffixed_path)
+            if stat is not None and not is_dir(stat):
+                return 0, suffixed_path, stat
+        return 404, None, None
+
+    path = f'{base_dir}/{relative_path}'
+
+    # If the requested path has a trailing slash, then it is intended
+    # to refer to a directory. We look for an index file inside the
+    # directory with the standard extensions, as well as 'index'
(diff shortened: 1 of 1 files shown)


examples/webserver/webserver.py
```diff
@@ -72,28 +72,36 @@ async def send_file(relative_path, request_headers):
...
         return Response(status_code=404, reason='Not Found')
-    absolute_path = f'{base_dir}/{relative_path}'
-
-    if absolute_path.endswith('/'):
-        absolute_path = f'{absolute_path}index.html'
-
-    try:
-        stat = await fs_async.stat(absolute_path)
-    except:
+    absolute_path = f'{base_dir}/{relative_path}'.rstrip('/')
```
wom-bat, line 78: We have had cases before where `http://example.com/fred.html` (accessed via `http://example.com/fred`and `http://example.com/fred/` are different URLs.
Please don't strip the trailing slash.
I suggest instead, 
```
if relative_path.endswith('/'):
    relative_path = relative_path + "index"
suffices = [ '', '.html', '.htm', '.xhtml' ]
for suffix in suffices:
  try stat = await fs_async.stat(relative_path + suffix)
  ...
  ```

This avoids pre-computing lots of paths, and creates only strings that are actually used.
If there's a trailing slash, it means this is a directory, and the index file inside that directory is whats wanted.  For our purposes at the moment that's always `index.html` but most webservers allow it to be configured.

**Review by wom-bat — CHANGES_REQUESTED** [review]
MInor changes ...

File: examples/webserver/webserver.py [hunk 2]
```diff
@@ -68,32 +68,87 @@ def format_http_date(timestamp):
...
-    try:
-        stat = await fs_async.stat(absolute_path)
-    except:
+    err, path, stat = await resolve(relative_path)
+    if err == 404:
         return Response(status_code=404, reason='Not Found')
-
-    if stat[0] & 0o170000 == 0o40000: # directory
-        return Response.redirect(f'/{relative_path}/', status_code=301)
+    elif err == 301:
```
wom-bat, line 143: Dont need 'elif' here as you've done a return in the previous branch; just an 'if'

wom-bat (changes requested):
One very minor change.  The rest looks good -- the restructuring works well.

wom-bat (approved):
Looks good.

PR #125, opened 2024-10-09
Fix directory website redirection

wom-bat:
When a URL referred to a directory, the webserver is meant to redirect to one with a trailing slash.  However it was returning a relative path, so, for example, https://sel4.org/Foundation/Summit redirected to https://sel4.org/Foundation/Summit/Foundation/summit/

Make it an absolute path.


wom-bat:
I thin this is the last code update before we can deploy the live seL4.systems to point here.

PR #132, opened 2024-10-09
ci: use multiple threads when compiling examples

Ivan-Velickovic:
CI is already taking now ~10 minutes which is too long when we only have 4 examples (we have to compile them in debug as well as release mode but still).


Ivan-Velickovic:
Maybe I'll revive this later but if we use multithreading with Makefiles, any error messages often become difficult to parse.

PR #136, opened 2024-10-14
Libmicrokitco update

dreamliner787-9:
This PR changes how coroutine stack pointers are passed into libmicrokitco's init function. 

Previously it is a variadic function, this is susceptible to programmer error as it is easy to pass in less or more pointers than required and it is not trivial to check how many arguments a variadic function received.

Now it leverages the type system so that the compiler can caught issues statically.


dreamliner787-9:
Whoops

PR #138, opened 2024-10-21
webserver: fix file handle leak from head request

JE-Archer:
Microdot treats a HEAD request as a special case of a GET request. Our
code looks the same for both: we fill out some headers and pass an
async file iterator to Microdot. However, in the case of HEAD request,
Microdot doesn't actually read from the file. The problem is that it
also doesn't close the file. Previously, we opened the file before
passing it to Microdot, so if Microdot didn't close the file then its
handle would leak. This commit changes our code so that we do not open
the file until Microdot actually tries to read from it (because in
that case it will also close it when done reading).


wom-bat (approved):
LGTM

PR #139, opened 2024-10-25
Add a deploy config option

Kswin01:
This PR adds a new option: `KITTY_CONFIG`. When this flag is set to `deploy`, the makefile will set the exec module to the `client/deploy.py` script, which in our example automatically runs the kitty script with a hardcoded server address, in our case our lab `fridge` server.


examples/kitty/kitty.mk
```diff
@@ -37,6 +37,13 @@ else
 $(error Unsupported MICROKIT_BOARD given)
 endif
 
+# If the KITTY_CONFIG is in deploy, then we will set the exec module
+# of Micropython to be the "deploy.py" file resident in the client directory.
+ifeq ($(strip $(KITTY_CONFIG)), deploy)
```
Ivan-Velickovic, line 42: Instead just have `DEPLOY` I think

examples/kitty/client/deploy.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2024, UNSW
+# SPDX-License-Identifier: BSD-2-Clause
+
+import kitty
+
+print("We are starting kitty.")
```
Ivan-Velickovic, line 6:
```suggestion
print("Deployment Kitty starting.")
```

PR #143, opened 2024-12-06
Adds virtualised filesystem example

dreamliner787-9:
This PR adds virtualised filesystem example into LionsOS. Currently, this example only supports `qemu_virt_aarch64`. 

See the README for more details: [https://github.com/au-ts/lionsos/blob/3d16d0f2b678ad1e113d5e48fe2bb2597e257cdc/examples/vfs/README.md](url)

Dependancies:
- [ ] Merge [https://github.com/au-ts/libvmm/pull/142](url)
- [ ] Repoint `dep/libvmm` to latest main


dreamliner787-9:
Switching to draft for now due to closure of [https://github.com/au-ts/libvmm/pull/142](url)

Things to fix:
- [ ] Make copy of liburing, similar to how lwip was handled.
- [ ] Copy FS UIO driver from libvmm into `components/fs/linux` (or similar).
- [ ] Fix vfs example build script to reflect the above.
- [ ] Make a physical board target for the vfs example.

Ivan-Velickovic:
> What is the rationale for vendoring liburing instead of adding it as a submodule? (I'm not sure we have a rationale for why we do this for lwip.)

It's really only *our* dependencies that should be submodules, since they update so often. Anything that in the long-term doesn't change much such as lwIP, MicroPython or liburing can just be copy-and-pasted.

Dealing with submodules really doesn't scale well. In general we'll likely have to find another method of getting dependency source code but for now I think we should minimise the amount of submodules.

lsf37:
Just pointing out here that this discussion has happened before in almost the same words and ended in using Google repo.

Ivan-Velickovic:
> Just pointing out here that this discussion has happened before in almost the same words and ended in using Google repo.

Google repo would be going backwards from where we are now the way I see it.

lsf37:
You're free to decide whatever you like, I'm just pointing out that I see history repeating. Both options suck. Monorepos also suck. All solutions to this problem that I have ever heard of suck, because ultimately what actually sucks are frequently changing dependencies. You can choose how much flexibility vs overhead you want, but that's pretty much it.

Ivan-Velickovic:
> You're free to decide whatever you like, I'm just pointing out that I see history repeating. Both options suck. Monorepos also suck. All solutions to this problem that I have ever heard of suck, because ultimately what actually sucks are frequently changing dependencies. You can choose how much flexibility vs overhead you want, but that's pretty much it.

You're probably right but the thing that I really don't like about both these solutions is that as a user of LionsOS I have somehow become responsible for managing the projects dependency. This is true for git submodules as well as the google repo tool.

examples/vfs/board/qemu_virt_aarch64/fs_driver_vm/linux
Ivan-Velickovic, line 1: Just putting this as a note to my self so I don't merge: we can't store any large artifacts in the repositories.

examples/vfs/board/qemu_virt_aarch64/fs_driver_vm/rootfs.cpio.gz
Ivan-Velickovic, line 1: Same here.

PR #146, opened 2025-01-10
Create FS server library

JE-Archer:
Implementing a LionsOS filesystem component involves both device and filesystem-related code that will be specific to the component as well as client-facing code that should be common to all LionsOS filesystems. This PR abstracts client-facing code from the NFS component into a FS server library and then employs the library in the NFS and FAT components.


lib/fs/server/memory.c
Ivan-Velickovic, line 1: Prefix with `fs_`

include/lions/fs/server.h
```diff
@@ -0,0 +1,26 @@
...
+int fd_free(fd_t fd);
+int fd_set_file(fd_t fd, void *file_handle);
+int fd_set_dir(fd_t fd, void *dir_handle);
+int fd_unset(fd_t fd);
+int fd_begin_op_file(fd_t fd, void **file_handle);
+int fd_begin_op_dir(fd_t fd, void **dir_handle);
+void fd_end_op(fd_t fd);
+
+void *fs_get_client_buffer(char *client_share, size_t client_share_size, fs_buffer_t buf);
+int fs_copy_client_path(char *dest, char *client_share, size_t client_share_size, fs_buffer_t buf);
```
Ivan-Velickovic, line 26:
```suggestion
/* Assumes dest is at least the size of buf.size + 1. */
int fs_copy_client_path(char *dest, char *client_share, size_t client_share_size, fs_buffer_t buf);
```

PR #148, opened 2025-01-13
Add make snippet for MicroPython

JE-Archer:


examples/webserver/webserver.mk
```diff
@@ -74,20 +74,14 @@ ${CHECK_FLAGS_BOARD_MD5}:
...
-			CONFIG_INCLUDE=$(abspath $(CONFIG_INCLUDE)) \
-			FROZEN_MANIFEST=$(abspath ./manifest.py) \
-			EXEC_MODULE=webserver.py
+MICROPYTHON_LIBMATH := $(LIBMATH)
+MICROPYTHON_CONFIG_INCLUDE := $(CONFIG_INCLUDE)
+MICROPYTHON_EXEC_MODULE := webserver.py
+MICROPYTHON_FROZEN_MANIFEST := manifest.py
+include $(LIONSOS)/components/micropython/micropython.mk
+
+manifest.py: webserver.py config.py
```
wom-bat, line 83: manifest.py doesn't get updated if webserver.py and config.py change.  So it's not a dependency.  you already have micropython.elf depending on these two files so if they change the recursive make will happen.

components/micropython/micropython.mk
```diff
@@ -0,0 +1,44 @@
...
+#	MICROPYTHON_EXEC_MODULE
+#	MICROPYTHON_ENABLE_I2C
+#	MICROPYTHON_ENABLE_FRAMEBUFFER
+#	MICROPYTHON_ENABLE_VFS_STDIO
+#	MICROPYTHON_ENABLE_SERIAL_STDIO
+
+MICROPYTHON_DIR := $(realpath $(dir $(lastword $(MAKEFILE_LIST))))
+
+micropython.elf: FORCE mpy-cross ${LIONSOS}/dep/libmicrokitco/Makefile $(MICROPYTHON_FROZEN_MANIFEST) $(MICROPYTHON_EXEC_MODULE)
+	make -C $(MICROPYTHON_DIR) \
```
wom-bat, line 24: use `${MAKE}` not `make`
Because 'FORCE' is on it'll always be executed, so you need no other dependencies, unless you want to make sure they're built before executing this line.

examples/kitty/kitty.mk
```diff
@@ -145,28 +146,17 @@ vmm.elf: ${VMM_OBJS} libvmm.a
...
+MICROPYTHON_ENABLE_I2C := 1
+MICROPYTHON_ENABLE_FRAMEBUFFER := 1
+include $(MICROPYTHON)/micropython.mk
 
 config.py: ${KITTY_DIR}/board/$(MICROKIT_BOARD)/config.py
 	cp $< $@
 
-manifest.py: ${KITTY_DIR}/manifest.py
+manifest.py: ${KITTY_DIR}/manifest.py kitty.py pn532.py font_height50.py font_height35.py writer.py config.py
 	cp $< $@
```
wom-bat, line 160: manifest.py isn't changed if kitty.py etc., change --- they're not dependencies.
The cp line here will copy the most recent of them to `manifest.py` which is not what you want.

examples/webserver/webserver.mk
```diff
@@ -74,20 +74,14 @@ ${CHECK_FLAGS_BOARD_MD5}:
...
+webserver.py: $(MICRODOT) config.py
```
wom-bat, line 84: Likewise, webserver.py doesn't change if config.py changes.

components/micropython/micropython.mk
```diff
@@ -0,0 +1,44 @@
...
+#	MICROPYTHON_FROZEN_MANIFEST
+#	MICROPYTHON_EXEC_MODULE
+#	MICROPYTHON_ENABLE_I2C
+#	MICROPYTHON_ENABLE_FRAMEBUFFER
+#	MICROPYTHON_ENABLE_VFS_STDIO
+#	MICROPYTHON_ENABLE_SERIAL_STDIO
+
+MICROPYTHON_DIR := $(realpath $(dir $(lastword $(MAKEFILE_LIST))))
+
+micropython.elf: FORCE mpy-cross ${LIONSOS}/dep/libmicrokitco/Makefile $(MICROPYTHON_FROZEN_MANIFEST) $(MICROPYTHON_EXEC_MODULE)
```
wom-bat, line 23: It'd be better to depend not on the Makefile for libmicrokitco but on the generated library I think, if that's possible.  We'll be moving to a snippet for libmicrokitco eventually I expect, so the Makefile will go away.
Right now, the Micropython component Makefile builds libmicrokitco, but I think we want to change that.

examples/kitty/kitty.mk
wom-bat: Yes you're right.

examples/webserver/webserver.mk
wom-bat: You can fix this by adding:
```
# Micropython includes frozen versions of various python sources;
# note them here
micropython.elf: manifest.py webserver.py config.py
```

wom-bat (changes requested):

PR #158, opened 2025-03-26
various CI + flake fixes

midnightveil:
1. remove a bunch of warnings from nix quirks
2. re-add nix examples to CI
3. ~~compiler faster by using the power of *multiprocessors*~~ nevermind
4. make nix not use dependencies from the ambient environment


Ivan-Velickovic:
Thanks a lot for looking into that!!

PR #165, opened 2025-05-29
Firewall: Add in ICMP module for generating ICMP packets.

Kswin01:
Adding an ICMP module for sending Destination Unreachable packets back to a source.

~DO NOT MERGE: There is a checksum issue present in this system where a ping request will show on the internal network as having a bad checksum of `0x1c1d`.~

### DIFF
--- a/examples/firewall/arp/arp_requester.c
+++ b/examples/firewall/arp/arp_requester.c
@@ -267,6 +267,54 @@ static uint16_t arp_table_flush(void) {
     return flushed;
 }
 
+/* Returns the number of ARP's still requiring a retry. */
+int process_retries(uint64_t time)
+{
+    int pending_pkts = 0;
+    // Loop over hashmap, and check if we have any pending retries.
+    for (int i = 0; i < 100; i++) {
+        if (arp_table->used[i] == 1 ) {
+            if (arp_table->entries[i].value.state == pending) {
+                if (arp_table->entries[i].value.num_retries >= ARP_MAX_RETRIES) {
+                    // ARP table has been expired
+                    arp_table->entries[i].value.state = negative;
+                    // Generate an ARP response packet for the router.
+                    arp_enqueue_response(arp_queues[arp_table->entries[i].value.client], arp_table->entries[i].key, empty_mac, false);
+                } else {
+                    // Resend the ARP request out to the network
+                    /* Generate ARP request */
+                    pending_pkts++;
+
+                    net_buff_desc_t buffer = {};
+                    int err = net_dequeue_free(&tx_queue, &buffer);
+                    if (err) {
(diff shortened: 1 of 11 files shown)


Kswin01:
Checksumming issue is fixed, this PR should be OK to merge after review.

examples/firewall/arp/arp_requester.c
```diff
@@ -47,7 +47,7 @@ static bool transmitted;
 static bool notify_client[FW_NUM_ARP_REQUESTER_CLIENTS] = {false};
 
 #define ARP_MAX_RETRIES 5               /* How many times the ARP requester will send out an ARP request. */
-#define ARP_RETRY_TIMER_S 5             /* How often to retry an ARP request, in seconds. */
+#define ARP_RETRY_TIMER_S 1             /* How often to retry an ARP request, in seconds. */
```
Courtney3141, line 50: Is this more accurate to realistic timing? Makes sense.

examples/firewall/icmp_module.c
```diff
@@ -0,0 +1,157 @@
...
+    net_buffers_init(&state.net1_queue, 0);
+
+    net_queue_init(&state.net2_queue, net2_config.tx.free_queue.vaddr, net2_config.tx.active_queue.vaddr,
+        net2_config.tx.num_buffers);
+    net_buffers_init(&state.net2_queue, 0);
+}
+
+void notified(microkit_channel ch)
+{
+    if (ch == icmp_config.router1_conn.ch) {
```
Courtney3141, line 148: Is it a little confusing here that router 1 is network 2 and router 2 is network 1? Is it possible to stick to the convention of external/internal (even if we end up changing that convention later)?

examples/firewall/meta.py
```diff
@@ -39,6 +39,9 @@ def round_up_to_Page(region_size: int) -> int:
 arp_queue_capacity = 512
 arp_queue_region_size = round_up_to_Page(2 * (4 + 16 * arp_queue_capacity) + 4)
 
+icmp_queue_capacity = 128
+icmp_queue_region_size = round_up_to_Page(312 * icmp_queue_capacity)
```
Courtney3141: Is this correct? Maybe it is, but just to be a bit clearer could you change it so that it shows which part is attributed to the icmp requests and which part is the head and tail? E.g. the queue is head, tail and request array, which would be (in bytes):
2 + 2 + 4 (padding)  + capacity * sizeof(request)

Courtney3141 (approved):

examples/firewall/arp/arp_requester.c
Kswin01: Usually a configurable timing, but IBM for instance allow users to enter a value between 500-5000 milliseconds, but they default to 500. One seconds is a good timeout for testing our system imo.

PR #166, opened 2025-06-04
Added support for new I2C protocol

omeh-a:
This PR adds support for the new sDDF i2c protocol from PR https://github.com/au-ts/sddf/pull/434. I've observed that there may be some fragility in our existing link to micropython and I have not altered this greatly yet, but we should be ready to make further changes in the future if usage changes.


Ivan-Velickovic:
Replaced by https://github.com/au-ts/lionsos/pull/233.

PR #167, opened 2025-07-09
Callum firewall

cazb2:
Changes to the routing table data structure

### DIFF
--- a/examples/firewall/webserver/modfirewall.c
+++ b/examples/firewall/webserver/modfirewall.c
@@ -89,8 +89,7 @@ fw_os_err_t filter_err_to_os_err(fw_filter_err_t filter_err) {
 extern fw_webserver_config_t fw_config;
 
 typedef struct fw_webserver_interface_state {
-    fw_routing_table_t routing_table;
-    uint16_t num_routes;
+    fw_routing_table_t *routing_table;
   
     fw_filter_state_t filter_states[FW_MAX_FILTERS];
     uint16_t num_rules[FW_MAX_FILTERS];
@@ -110,10 +109,7 @@ void fw_webserver_init(void) {
     sddf_memcpy(webserver_state.mac_addr, fw_config.interfaces[fw_config.interface].mac_addr, ETH_HWADDR_LEN);
     
     for (uint8_t i = 0; i < FW_NUM_INTERFACES; i++) {
-        webserver_state.interfaces[i].routing_table.entries = fw_config.interfaces[i].router.routing_table.vaddr;
-        webserver_state.interfaces[i].routing_table.capacity = fw_config.interfaces[i].router.routing_table_capacity;
-        /* Hardcode pre-existing interface route. */
-        webserver_state.interfaces[i].num_routes = 1;
+        webserver_state.interfaces[i].routing_table = fw_config.interfaces[i].router.routing_table.vaddr;
 
         for (uint8_t j = 0; j < fw_config.interfaces[i].num_filters; j++) {
             fw_filter_state_init(&webserver_state.interfaces[i].filter_states[j],
@@ -122,9 +118,6 @@ void fw_webserver_init(void) {
                                     fw_config.interfaces[i].filters[j].default_action);
         }
     }
-
-  /* Currently harcode pre-existing internal route to webserver. */
-  webserver_state.interfaces[FW_INTERNAL_INTERFACE_ID].num_routes += 1;
 }
 
 /* Get MAC address for network interface */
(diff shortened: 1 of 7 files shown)


include/lions/firewall/common.h
```diff
@@ -2,6 +2,7 @@
 
 #include <stdint.h>
 #include <sddf/util/util.h>
+#include <string.h>
```
Courtney3141: I try to reserve this common.h file for things that are used by basically *every* firewall component. Since this array shift function is only used by routers and filters, could you please move it into another header file? Possibly something like "consecutive_array.h" or a better name if you can think of one.

include/lions/firewall/routing.h
```diff
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <stdint.h>
```
Courtney3141, line 3: Sorry I possibly accidentally removed this... Could you add this back in

include/lions/firewall/routing.h
```diff
@@ -295,19 +292,13 @@ static fw_routing_err_t fw_routing_table_add_route(fw_routing_table_t *table,
     /* Default routes must specify a next hop! */
     if ((subnet == 0) && (next_hop == FW_ROUTING_NONEXTHOP)) {
         return ROUTING_ERR_INVALID_ROUTE;
+    } else if (table->size >= table->capacity) {
+        return ROUTING_ERR_FULL;
     }
 
-    fw_routing_entry_t *empty_slot = NULL;
-    for (uint16_t i = 0; i < table->capacity; i++) {
+    for (uint16_t i = 0; i < table->size; i++) {
```
Courtney3141, line 301: If we are changing to ++i (which is fine), keep consistant

Courtney3141: Possibly I deleted your change here! (sorry)

include/lions/firewall/routing.h
```diff
@@ -326,10 +317,12 @@ static fw_routing_err_t fw_routing_table_add_route(fw_routing_table_t *table,
         }
     }
 
-    if (empty_slot == NULL) {
+    if (table->size == table->capacity) {
```
Courtney3141: Since we check at the beginning of the function whether `size >= capacity` this check should no longer be needed (relic of old implementation)

include/lions/firewall/routing.h
```diff
@@ -326,10 +317,12 @@ static fw_routing_err_t fw_routing_table_add_route(fw_routing_table_t *table,
...
         return ROUTING_ERR_FULL;
     }
 
+    fw_routing_entry_t *empty_slot = table->entries + table->size;
```
Courtney3141, line 322: Although it may not be so important in this case (as there is only one reader other than us), it's always important that "size" is only incremented *after* the entry becomes valid. I.e if size is incremented, that means that index size - 1 is assumed to be valid, which is not true unless the fields have finished being filled. So increment size *after* the other fields are set. (I understand the previous code may have been incorrect here too!)

include/lions/firewall/array_functions.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <stdint.h>
+#include <sddf/util/util.h>
+#include <string.h>
+
+static void generic_array_shift(void* array, uint32_t entry_size, uint32_t array_len, uint32_t start_point) {
```
Courtney3141: Could you please change the name of `start_point` to something a bit more intuitive, possibly `index_to_remove`?

include/lions/firewall/common.h
```diff
@@ -2,6 +2,7 @@
```
Courtney3141: Please drop changes of this file (string.h should not be needed)

examples/firewall/webserver/modfirewall.c
```diff
@@ -269,10 +259,10 @@ STATIC mp_obj_t route_get_nth(mp_obj_t interface_idx_in,
 
     uint16_t valid_entries = 0;
     for (uint16_t i = 0;
-        i < webserver_state.interfaces[interface_idx].routing_table.capacity; i++) {
+        i < webserver_state.interfaces[interface_idx].routing_table->size; i++) {
```
Courtney3141: Totally forgot that now that we know the valid entries are consecutive, we no longer need to loop through the structure to find the `n`th entry! Can you please change this so that once we have verified `route_idx` is valid, just return the `route_idx`th entry values (i.e. `routing_table->entries + route_idx`.)

Courtney3141 (approved):

PR #168, opened 2025-07-16
Firewall filter data structure changes

cazb2:
Modified the filter rules data structures to be a container encapsulating size and capacity.
Modified the way rules get allocated ids and therefore had to change the modfirewall webserver functions.
Added a bitmap to track the ids being used.

### DIFF
--- a/examples/firewall/meta.py
+++ b/examples/firewall/meta.py
@@ -504,7 +504,7 @@ def generate(sdf_file: str, output_dir: str, dtb: DeviceTree, iotgate_idx: int):
             # create bitmap
             mr = MemoryRegion(sdf, "rules_id_bitmap" + "_" + filter_pd.name, filter_rule_bitmap_region_size);
             sdf.add_mr(mr);
-            bitmap_region = fw_region(filter_pd, mr, "rw",filter_rule_bitmap_region_size);
+            rule_bitmap_region = fw_region(filter_pd, mr, "rw",filter_rule_bitmap_region_size);
 
             # Create rule region
             filter_rules = fw_shared_region(filter_pd, webserver, "rw", "r", "filter_rules", filter_rule_region_size)
@@ -541,7 +541,7 @@ def generate(sdf_file: str, output_dir: str, dtb: DeviceTree, iotgate_idx: int):
                 filter_webserver_config,
                 None,
                 None,
-                bitmap_region
+                rule_bitmap_region
             )
 
             network["configs"][router].filters.append((filter_router_conn[1]))
--- a/include/lions/firewall/filter.h
+++ b/include/lions/firewall/filter.h
@@ -9,6 +9,7 @@
 
 #define MOD64 0x3F
 #define BITMAP_BLOCK_SIZE 64
+#define DEFAULT_RULE 0
 
 typedef enum {
     FILTER_ERR_OKAY = 0,   /* No error */
@@ -106,7 +107,7 @@ static fw_filter_err_t rules_reserve_id(fw_rule_id_bitmap_t *bitmap, uint16_t ru
 
 /* Sets the bit for the inputted id rule to 0, will never free the default rule id (0) */
 static void rules_free_id(fw_rule_id_bitmap_t *bitmap, uint16_t id) {
-    if (id == 0) {
+    if (id == DEFAULT_RULE) {
         return;
     }
     uint16_t block = id / BITMAP_BLOCK_SIZE;
@@ -118,8 +119,8 @@ typedef struct fw_filter_state {
     fw_rule_table_t *rule_table; /* Container for firewall rules */
     fw_rule_id_bitmap_t *rule_id_bitmap; /* Functions for interacting with it assumes 0*/
 
-    fw_instances_table_t *internal_instances_container /* Instances for other filter to check */;
-    fw_instances_table_t *external_instances_container; /* Instances for this filter to check against */
+    fw_instances_table_t *internal_instances_table /* Instances for other filter to check */;
+    fw_instances_table_t *external_instances_table; /* Instances for this filter to check against */
 
     fw_action_t default_action;
 } fw_filter_state_t;
@@ -172,10 +173,10 @@ static void fw_filter_state_init(fw_filter_state_t *state,
         sddf_printf("ERROR ON FILTER BIT MAP CREATION\n");
     }
 
-    state->internal_instances_container = (fw_instances_table_t *)internal_instances;
-    state->internal_instances_container->capacity = instances_capacity;
+    state->internal_instances_table = (fw_instances_table_t *)internal_instances;
+    state->internal_instances_table->capacity = instances_capacity;
 
-    state->external_instances_container = (fw_instances_table_t *)external_instances;
+    state->external_instances_table = (fw_instances_table_t *)external_instances;
 
     state->default_action = default_action;
 }
@@ -192,6 +193,10 @@ static fw_filter_err_t fw_filter_add_rule(fw_filter_state_t *state,
                                           fw_action_t action,
                                           uint16_t *rule_id)
 {
+    if (state->rule_table->size >= state->rule_table->capacity) {
+        return FILTER_ERR_FULL;
+    }
+
     for (uint16_t i = 0; i < state->rule_table->size ; i++) {
         fw_rule_t *rule = (fw_rule_t *)(state->rule_table->rules + i);
 
@@ -235,10 +240,6 @@ static fw_filter_err_t fw_filter_add_rule(fw_filter_state_t *state,
         }
     }
 
-    if (state->rule_table->size >= state->rule_table->capacity) {
-        return FILTER_ERR_FULL;
-    }
-    
     fw_rule_t *empty_slot = state->rule_table->rules + state->rule_table->size;
     empty_slot->src_ip = subnet_mask(src_subnet) & src_ip;
     empty_slot->src_port = src_port;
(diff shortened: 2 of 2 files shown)


include/lions/firewall/config.h
```diff
@@ -109,6 +109,7 @@ typedef struct fw_filter_config {
     fw_webserver_filter_config_t webserver;
     region_resource_t internal_instances;
     region_resource_t external_instances;
+    region_resource_t rules_id_bitmap;
```
Courtney3141: I think `rule_id_bitmap` is better grammar.

include/lions/firewall/filter.h
```diff
@@ -39,36 +40,83 @@ static const char *fw_filter_action_str[] = {
...
-    bool valid;
     uint32_t src_ip;
     uint16_t src_port;
     uint32_t dst_ip;
     uint16_t dst_port;
     uint8_t src_subnet; /* ip any is encoded as subnet mask 0 */
     uint8_t dst_subnet;
-    bool src_port_any;
-    bool dst_port_any;
+    bool src_port_any; // can we delete this
```
Courtney3141: No since there is no valid way to represent that the rule applies to *any* port number (rather than a specific one) through the other members of the rule struct. (Unless you can think of one)

include/lions/firewall/filter.h
```diff
@@ -39,36 +40,83 @@ static const char *fw_filter_action_str[] = {
...
     fw_action_t action;
+    uint16_t rule_id;
 } fw_rule_t;
 
 /* These entries are to inform the other filter component that a connection
 has been established, and that traffic should now be allowed back */
 typedef struct fw_instance {
-    bool valid;
-    bool default_rule; /* indicates that this is an instance of the default rule */
+    uint8_t flags; // lsb used for default_rule flag
```
Courtney3141: I thought we were going to give the default rule a fixed number like 0, so it can be encoded in the `rule_id` field.

include/lions/firewall/filter.h
```diff
@@ -39,36 +40,83 @@ static const char *fw_filter_action_str[] = {
...
     uint16_t rule_id; /* index of the rule that this is an instance of. Book-keeping for instance creator. */
     uint32_t src_ip; /* ip of connection target */
     uint16_t src_port; /* port of connection target */
     uint32_t dst_ip; /* ip of connection creator */
     uint16_t dst_port; /* port of connection creator */
 } fw_instance_t;
 
+typedef struct fw_instances_container {
```
Courtney3141: Let's try to stick to a naming convention with entries/tables holding them. In `routing.h` we have `routing_entry` and `routing_table`. I think it makes sense to keep the names `fw_rule_t` and `fw_instance_t`, but maybe the container data structures should be `fw_rule_table_t` and `fw_instance_table_t`

include/lions/firewall/filter.h
```diff
@@ -39,36 +40,83 @@ static const char *fw_filter_action_str[] = {
...
+    fw_instance_t instances[];
+} fw_instances_container_t;
+
+typedef struct fw_rules {
+    uint16_t capacity;
+    uint16_t size;
+    fw_rule_t rules[];
+} fw_rules_container_t;
+
+typedef struct fw_rules_bitmap {
```
Courtney3141: Lets be more specific and call this `fw_rule_id_bitmap_t` for clarity.

include/lions/firewall/filter.h
```diff
@@ -103,17 +151,27 @@ typedef enum {
...
                                  fw_action_t default_action)
 {
-    state->rules = (fw_rule_t *)rules;
-    state->rules_capacity = rules_capacity;
-    state->internal_instances = (fw_instance_t *)internal_instances;
-    state->external_instances = (fw_instance_t *)external_instances;
-    state->instances_capacity = instances_capacity;
+    state->rules_container = (fw_rules_container_t *)rules;
+    state->rules_container->capacity = rules_capacity;
+    state->rules_container->size = 0;
```
Courtney3141: The filter state is a global variable in the data section so is by default zero initialised.

include/lions/firewall/filter.h
```diff
@@ -180,11 +230,11 @@ static fw_filter_err_t fw_filter_add_rule(fw_filter_state_t *state,
         }
     }
 
-    if (empty_slot == NULL) {
+    if (state->rules_container->size >= state->rules_container->capacity) {
```
Courtney3141: This check is faster so should be performed first.

examples/firewall/meta.py
```diff
@@ -247,7 +251,8 @@ def generate(sdf_file: str, output_dir: str, dtb: DeviceTree, iotgate_idx: int):
     # Create serial subsystem
     common_pds.append(ProtectionDomain("serial_driver", "serial_driver.elf", priority=100))
     common_pds.append(ProtectionDomain("serial_virt_tx", "serial_virt_tx.elf", priority=99))
-    serial_system = Sddf.Serial(sdf, serial_node, common_pds[-2], common_pds[-1])
+    # serial_system = Sddf.Serial(sdf, serial_node, common_pds[-2], common_pds[-1])
+    serial_system = Sddf.Serial(sdf, serial_node, common_pds[-2], common_pds[-1], enable_color = False)
```
Courtney3141: Remove this

examples/firewall/meta.py
```diff
@@ -497,6 +502,11 @@ def generate(sdf_file: str, output_dir: str, dtb: DeviceTree, iotgate_idx: int):
             network["in_net"].add_client_with_copier(filter_pd, tx=False)
             network["configs"][in_virt].active_client_protocols.append(protocol)
 
+            # create bitmap
+            mr = MemoryRegion(sdf, "rules_id_bitmap" + "_" + filter_pd.name, filter_rule_bitmap_region_size);
+            sdf.add_mr(mr);
+            bitmap_region1 = fw_region(filter_pd, mr, "rw",filter_rule_bitmap_region_size);
```
Courtney3141: `bitmap_region1` --> `rule_bitmap_region`

include/lions/firewall/filter.h
```diff
@@ -39,36 +40,83 @@ static const char *fw_filter_action_str[] = {
...
+} fw_instances_container_t;
+
+typedef struct fw_rules {
+    uint16_t capacity;
+    uint16_t size;
+    fw_rule_t rules[];
+} fw_rules_container_t;
+
+typedef struct fw_rules_bitmap {
+    uint16_t capacity;
```
Courtney3141: capacity can be extracted from rules data structure and does not need to be stored separately.

include/lions/firewall/filter.h
```diff
@@ -39,36 +40,83 @@ static const char *fw_filter_action_str[] = {
...
+
+/* Functions for the bitmap*/
+/* Finds the first rule that is not in use, and marks it as used */
+uint16_t rules_reserve_id(fw_rules_bitmap_t *bitmap) {
+    uint16_t total_ids = bitmap->capacity * 64;
+    for (uint16_t i = 0; i < total_ids; i++) {
+        uint16_t id_to_check = (bitmap->last_allocated_rule_id + i) % total_ids;
+
+        uint16_t block_idx = id_to_check / 64;
+        uint8_t bit_pos = id_to_check & 0x3F;
```
Courtney3141: Possibly define these constants (`64/0x3F`) as macros.

include/lions/firewall/filter.h
```diff
@@ -39,36 +40,83 @@ static const char *fw_filter_action_str[] = {
...
+        uint64_t mask = 1ULL << bit_pos;
+
+        if (!(bitmap->id_bitmap[block_idx] & mask)) {
+            bitmap->id_bitmap[block_idx] |= mask;
+            bitmap->last_allocated_rule_id = id_to_check;
+            return id_to_check;
+        }
+    }
+    /* This shouldn't happens since we only call this function if there is space in the rules[]*/
+    return 0;
```
Courtney3141: Return a firewall error type and modify a provided pointer for allocated rule.

include/lions/firewall/filter.h
```diff
@@ -39,36 +40,83 @@ static const char *fw_filter_action_str[] = {
...
+}
+
+/* Sets the bit for the inputted id rule to 0 */
+void rules_free_id(fw_rules_bitmap_t *bitmap, uint16_t id) {
```
Courtney3141: `static`

include/lions/firewall/filter.h
```diff
@@ -329,23 +368,20 @@ static fw_action_t fw_filter_find_action(fw_filter_state_t *state,
...
+        if (default_rule && (instance->flags & 0b1)) {
             continue;
         }
         
         if (!default_rule && (rule_id != instance->rule_id)) {
             continue;
         }
-
-        instance->valid = false;
+        state->internal_instances_container->instances[i] = state->internal_instances_container->instances[state->internal_instances_container->size - 1];
```
Courtney3141: Nice idea but this `for` loop skips the newly relocated instance which is now sitting at `i`

include/lions/firewall/filter.h
```diff
@@ -363,20 +399,29 @@ static fw_filter_err_t fw_filter_update_default_action(fw_filter_state_t *state,
     return FILTER_ERR_OKAY;
 }
 
-static fw_filter_err_t fw_filter_remove_rule(fw_filter_state_t *state, uint8_t rule_id)
+static fw_filter_err_t fw_filter_remove_rule(fw_filter_state_t *state, uint16_t rule_id)
 {
-    fw_rule_t *rule = state->rules + rule_id;
-    if (rule_id >= state->rules_capacity || !rule->valid) {
+    fw_rule_t *rule = NULL;
```
Courtney3141, line 417: add check for default rule ID

include/lions/firewall/filter.h
```diff
@@ -363,20 +399,29 @@ static fw_filter_err_t fw_filter_update_default_action(fw_filter_state_t *state,
...
+    for (uint16_t i = 0; i < state->rules_container->size; i++) {
+        if (state->rules_container->rules[i].rule_id == rule_id) {
+            rule = state->rules_container->rules + i;
```
Courtney3141: should be a `break` here.

include/lions/firewall/filter.h
cazb2: I don't think I understand what you mean here.  What check are you referring to as the faster one?

include/lions/firewall/filter.h
cazb2: had a cheeky idea for this but you might have strong feelings about it

include/lions/firewall/filter.h
Courtney3141: I meant check whether the 'container' is full *before* checking if there is a clash/duplicate since the fullness check is faster.

include/lions/firewall/filter.h
Courtney3141: I guess you'll have to ask me haha

PR #170, opened 2025-08-08
examples.sh: re-enable fileio

Ivan-Velickovic:


Ivan-Velickovic:
Will return to this after https://github.com/au-ts/lionsos/pull/171, it should solve the root problem.

PR #171, opened 2025-08-21
WIP: clean up libc and use clang for everything

dumsum:
Early proof of concept for cleanup. Feedback welcome.

- Remove dependency on packaged toolchain libc and link against our musllibc fork instead.
- Move "posix" layer out of the nfs component
- Use clang for almost everything. (mpy-cross excepted)
- Not sure about linking compiler-rt instead of libgcc (also I expect this only to work in Nix at the moment depending on where the library is located relative to clang in other environments). It seems we still need this to support 128-bit floating point arithmetic, perhaps we could vendor those functions instead.
- Also some small changes to libnfs, be sure to update the submodule


Ivan-Velickovic:
@dumsum I will merge https://github.com/au-ts/lionsos/pull/172 first probably which means it would be a good idea to rebase your branch on that. I don't know if there will be any conflicts but things like `components/micropython/newlibc_stubs.c` should be removed, it was added to bridge the gap between the latest sDDF with your custom libc changes and your PR here.

Ivan-Velickovic:
One more thing to come back to, `LWIP_DNS_SECURE` in `components/micropython/lwip_include/lwipopts.h` can be removed once we have `rand`.

Ivan-Velickovic:
Instead of `lib/fp` I would do `lib/compiler_rt` with either a sub-dir of `fp/` or just putting everything directly in `compiler_rt`. Just to make it obvious what it's used for.

Ivan-Velickovic:
Hmm, strange. I removed the step in the Linux CI that manually downloads `aarch64-none-elf-gcc` since we shouldn't need it anymore but now that run is failing with a build error for musllibc. Although I can't see what specifically is going wrong from the logs...

Ivan-Velickovic:
Found it:
```
aarch64-none-elf-ar: No such file or directory
```

dumsum:
Rebased on https://github.com/au-ts/lionsos/pull/173

include/lions/util.h
Ivan-Velickovic, line 1: For this one I would like to keep inside nfs, this `dlogp` stuff is not really used anywhere else.

Ivan-Velickovic: Nevermind, I am wrong.

PR #173, opened 2025-08-28
Add firewall example to LionsOS

Courtney3141:
This PR adds the LionsOS Firewall example to main. Along with this PR, over the next few days I will additionally be adding:
- Documentation to the LionsOS website
- Issue and missing feature tracking to git issues

Currently, the firewall offers a collection of rudimentary functionalities, summarised below:

## General
- Currently supports 2 network interfaces (external and internal)
- Traffic can from one interface to the other, as well from the internal interface to the webserver
- Traffic types supported:
    - ARP: Responses for the firewall, requests for outgoing traffic
    - ICMP: forwarding and transmitting `destination unreachable` packets
    - UDP & TCP: forwarding
- Most components (with the exception of the webserver and ICMP module) are duplicated per network interface 

## ARP
- The ARP requester component sends out ARP requests for IPs the system wishes to forward to
    - If a reply is received an ARP entry is inserted into a cache which is reused by the system until a periodic flush occurs
    - If a reply is not received after a configurable time interval (default 1 second), a duplicate request is send (up to 5 re-tries)

## Filters
- Filter components decide whether or not traffic is permitted to be forwarded through the firewall. Every packet received is directed by the virtualiser to the relevant filter (currently ICMP, UDP and TCP).
- Each packet is checked against the current rules table. If a match is not found the default rule is used. Currently the following options for actions are:
    - Allow
    - Drop
    - Connect (which allows both the packet *and* return traffic, requiring shared state between the filters)
- Rules can be updated at runtime using the webserver

## Routing
- If traffic is permitted by a filter, it is passed to the routing component which is responsible for updating the destination MAC address and transmitting the packet out the NIC.
- First, the routing table will be consulted to determine which NIC the packet should be transmitted through (currently only 1 possible NIC or internally to the webserver). 
- Routing table routes can be updated using the Micropython webserver UI.
- Second, the ARP table is checked for an existing entry for the destination IP. If one is found, the packet is updated and transmitted. If there is no match, the packet will be temporarily stored and the ARP requester will contacted to request the MAC. If the request comes back successfully, the packet will be updated and transmitted, else the packet will be dropped and the ICMP module will be notified.

## ICMP Module
- The ICMP module handles the sending of `destination unreachable` packets to senders if the firewall cannot find a route to the destination.
- Both routing components have a connection to the module.
- In the future, the module will be expanded to handle a larger class of ICMP packets.

## Webserver
- The Micropython webserver provides a web UI accessible from the internal network which allows filter rules and routing tables routes to be updated at runtime.

## Building
- Due to the high complexity of the system and the many connections between components, building the system without the use of additional python scripting is particularly challenging and easy to get wrong. Thus the firewall utilises a collection of newly developed scripts and techniques which automate much of this process.
- This can mostly be seen in the `sdfgen_helper.py` script which is automatically invoked with `make` and generates additional python modules which are used in `metaprogram.py`.
- These scripts and techniques will be expanded upon in the LionsOS docs.

## Performance
Performance is not yet a major focus of this project, but as a milestone I tested ping latency in both directions:

External --> Internal
PING 192.168.1.2 (192.168.1.2) 56(84) bytes of data.
64 bytes from 192.168.1.2: icmp_seq=1 ttl=63 time=0.967 ms
64 bytes from 192.168.1.2: icmp_seq=2 ttl=63 time=0.855 ms
64 bytes from 192.168.1.2: icmp_seq=3 ttl=63 time=1.04 ms
64 bytes from 192.168.1.2: icmp_seq=4 ttl=63 time=1.05 ms
64 bytes from 192.168.1.2: icmp_seq=5 ttl=63 time=0.771 ms
64 bytes from 192.168.1.2: icmp_seq=6 ttl=63 time=0.460 ms
64 bytes from 192.168.1.2: icmp_seq=7 ttl=63 time=0.986 ms
64 bytes from 192.168.1.2: icmp_seq=8 ttl=63 time=1.21 ms
64 bytes from 192.168.1.2: icmp_seq=9 ttl=63 time=1.16 ms
64 bytes from 192.168.1.2: icmp_seq=10 ttl=63 time=0.866 ms
64 bytes from 192.168.1.2: icmp_seq=11 ttl=63 time=0.774 ms
64 bytes from 192.168.1.2: icmp_seq=12 ttl=63 time=0.654 ms

Internal --> External
PING 172.16.0.2 (172.16.0.2) 56(84) bytes of data.
64 bytes from 172.16.0.2: icmp_seq=1 ttl=63 time=0.484 ms
64 bytes from 172.16.0.2: icmp_seq=2 ttl=63 time=0.851 ms
64 bytes from 172.16.0.2: icmp_seq=3 ttl=63 time=0.991 ms
64 bytes from 172.16.0.2: icmp_seq=4 ttl=63 time=0.990 ms
64 bytes from 172.16.0.2: icmp_seq=5 ttl=63 time=0.747 ms
64 bytes from 172.16.0.2: icmp_seq=6 ttl=63 time=0.578 ms
64 bytes from 172.16.0.2: icmp_seq=7 ttl=63 time=0.806 ms
64 bytes from 172.16.0.2: icmp_seq=8 ttl=63 time=1.07 ms
64 bytes from 172.16.0.2: icmp_seq=9 ttl=63 time=0.780 ms
64 bytes from 172.16.0.2: icmp_seq=10 ttl=63 time=0.740 ms
64 bytes from 172.16.0.2: icmp_seq=11 ttl=63 time=0.715 ms
64 bytes from 172.16.0.2: icmp_seq=12 ttl=63 time=0.777 ms

Note there is not currently a Microkit config for `benchmark` or `release` for the `iotgate`, so these tests were performed in `debug` with debug printing turned off.

### DIFF
--- a/examples/firewall/firewall.mk
+++ b/examples/firewall/firewall.mk
@@ -1,0 +1,242 @@
+# Makefile for firewall.
+#
+# Copyright 2025, UNSW
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+# This makefile will be copied into the Build directory and used from there.
+#
+BOARD_DIR := $(MICROKIT_SDK)/board/$(MICROKIT_BOARD)/$(MICROKIT_CONFIG)
+ARCH := $(shell grep 'CONFIG_SEL4_ARCH  ' $(BOARD_DIR)/include/kernel/gen_config.h | cut -d' ' -f4)
+SDDF := $(LIONSOS)/dep/sddf
+
+ifeq ($(strip $(MICROKIT_BOARD)), imx8mp_evk)
+	ETH_DRIV_DIR0 := imx
+	ETH_DRIV_DIR1 := dwmac-5.10a
+	SERIAL_DRIV_DIR := imx
+	TIMER_DRV_DIR := imx
+	CPU := cortex-a53
+else
+$(error Unsupported MICROKIT_BOARD given)
+endif
+
+TOOLCHAIN := clang
+CC := clang
+LD := ld.lld
+AR := llvm-ar
+RANLIB := llvm-ranlib
+OBJCOPY := llvm-objcopy
+OBJDUMP := llvm-objdump
+MICROKIT_TOOL ?= $(MICROKIT_SDK)/bin/microkit
+PYTHON ?= python3
+DTC := dtc
+
+ifeq ($(ARCH),aarch64)
+	CFLAGS_ARCH := -mcpu=$(CPU)
+	TARGET := aarch64-none-elf
+else ifeq ($(ARCH),riscv64)
+	CFLAGS_ARCH := -march=rv64imafdc
+	TARGET := riscv64-none-elf
+else
+$(error Unsupported ARCH given)
+endif
+
+ifeq ($(strip $(TOOLCHAIN)), clang)
+	CFLAGS_ARCH += -target $(TARGET)
+endif
+
+# Use sDDF custom libc for sDDF components
+SDDF_CUSTOM_LIBC := 1
+
+MICRODOT := $(LIONSOS)/dep/microdot/src
+FIREWALL_NET_COMPONENTS := $(FIREWALL_SRC_DIR)/net_components
+FIREWALL_FILTERS := $(FIREWALL_SRC_DIR)/filters
+FIREWALL_ICMP := $(FIREWALL_SRC_DIR)/icmp
+FIREWALL_ROUTING := $(FIREWALL_SRC_DIR)/routing
+FIREWALL_ARP := $(FIREWALL_SRC_DIR)/arp
+
+METAPROGRAM := $(FIREWALL_SRC_DIR)/meta.py
+DTS := $(SDDF)/dts/$(MICROKIT_BOARD).dts
+DTB := $(MICROKIT_BOARD).dtb
+
+SDFGEN_HELPER := $(FIREWALL_SRC_DIR)/sdfgen_helper.py
+# Macros needed by sdfgen helper to calculate config struct sizes
+SDFGEN_UNKOWN_MACROS := ETH_HWADDR_LEN=6 SDDF_NET_MAX_CLIENTS=64
+# Headers containing config structs and dependencies
+FIREWALL_CONFIG_HEADERS := $(SDDF)/include/sddf/resources/common.h \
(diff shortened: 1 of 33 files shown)


examples/firewall/firewall.mk
```diff
@@ -0,0 +1,242 @@
...
+		  $(SDDF)/serial/components/serial_components.mk
+
+include $(SDDF_MAKEFILES)
+include $(FIREWALL_NET_COMPONENTS)/firewall_network_components.mk
+
+$(DTB): $(DTS)
+	$(DTC) -q -I dts -O dtb $(DTS) > $(DTB)
+
+$(SYSTEM_FILE): $(METAPROGRAM) $(IMAGES) $(DTB) $(CHECK_FLAGS_BOARD_MD5)
+	$(PYTHON) $(SDFGEN_HELPER) --macros "$(SDFGEN_UNKOWN_MACROS)" --configs "$(FIREWALL_CONFIG_HEADERS)" --output $(FIREWALL_SRC_DIR)/config_structs.py
```
midnightveil: This config_structs.py appears to have been added to .gitignore and I can't see it in this repo anywhere.

Courtney3141: It is a build artifact and needs to be generated per build. But it has not been moved to the build directory.

examples/firewall/meta.py
```diff
@@ -0,0 +1,816 @@
...
+import subprocess
+from os import path
+from dataclasses import dataclass
+from typing import List
+from sdfgen import SystemDescription, Sddf, DeviceTree
+from ctypes import *
+from importlib.metadata import version
+import ipaddress
+
+assert version('sdfgen').split(".")[1] == "25", "Unexpected sdfgen version"
```
dumsum: Should this file be updated to sdfgen 26?

Ivan-Velickovic: Yes, there's been no breaking changes between 0.25 and 0.26 so fine to just bump it so it's consistent with all the other examples.

Ivan-Velickovic: My bad, there are were a couple…

dumsum: > My bad, there are were a couple…

Yeah all good, I think I figured it out! (everything compiles at least)

PR #197, opened 2025-09-04
examples/fileio: fix build fail

dreamliner787-9:
Fix failing build for `fileio` example due to a recent sDDF update that requires the partition schema to be specified in the `mkvirtdisk` script.


Ivan-Velickovic:
use GPT I reckon

PR #204, opened 2025-09-12
Libc fixes

dumsum:
With these changes:
- the musl build directory is now an order-only prerequisite to avoid wasting time with the configuration step each time a system is rebuilt
- micropython now uses musl's `printf` instead of its internal implementation (missed in the original libc changes)
- `lib_posix.mk` and `lib_compiler_rt.mk` are included at system level (alongside musl i.e. `libc.a`) - they should only need to be built once with components linking as needed.
    - Note similarity with https://github.com/au-ts/lionsos/commit/16ca6fa853dc817493bd1abe675af416a30afb68 
    - I feel this is a good first step in a more principled approach to our build system but keen for thoughts/feedback
    - I suppose an alternative would be to have each component build its own instance of these...


examples/kitty/meta.py
```diff
@@ -88,7 +88,7 @@ def generate(sdf_path: str, output_dir: str, dtb: DeviceTree):
     net_virt_rx = ProtectionDomain("net_virt_rx", "network_virt_rx.elf", priority=99)
     net_system = Sddf.Net(sdf, ethernet_node, ethernet_driver, net_virt_tx, net_virt_rx)
 
-    micropython = ProtectionDomain("micropython", "micropython.elf", priority=1, budget=20000)
+    micropython = ProtectionDomain("micropython", "micropython.elf", priority=1, budget=20000, stack_size=0x10000)
```
Ivan-Velickovic, line 91: Can you try with `stack_size=0x2000` and let me know if it's sufficient?

dumsum:
https://github.com/au-ts/lionsos/commit/6acf059465daf08fad184c41055c10206977ea61 is my attempt at cleaning up the dependencies.

- combines musl, posix and compiler_rt into one libc.a
- abstractly specifies libc dependencies (with LIONS_LIBC exported from libc.mk)
- many sDDF objects need libc headers (also libvmm). To ensure they are available, defines SDDF_LIBC_INCLUDE (corresponding to the build target for the libc include directory in LionsOS), which is added as build prerequisites in the sDDF/libvmm makefile snippets.

examples/kitty/meta.py
dumsum: In musl `vfprintf.c`, the `printf_core` function allocates about 2 pages of stack frame. So this would be on top of existing stack requirement.

As far as I can tell (limited testing -> spamming printf's), seems like nfs needs at least `0x5000` and micropython `0x3000`.

PR #208, opened 2025-09-17
Refactor common things out of Makefiles

wom-bat:
To ease adding new boards to projects, abstract out all the common bits into separate files.

Do the same for the toolchain bits.

I'm not sure whether the board descriptions should be here or in sDDF.

### DIFF
--- a/examples/firewall/firewall.mk
+++ b/examples/firewall/firewall.mk
@@ -6,44 +6,17 @@
 #
 # This makefile will be copied into the Build directory and used from there.
 #
-BOARD_DIR := $(MICROKIT_SDK)/board/$(MICROKIT_BOARD)/$(MICROKIT_CONFIG)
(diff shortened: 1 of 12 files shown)


examples/firewall/firewall.mk
```diff
@@ -55,41 +28,36 @@ MUSL_SRC := $(LIONSOS)/dep/musllibc
...
-CFLAGS := \
-	-mtune=$(CPU) \
-	-mstrict-align \
-	-ffreestanding \
-	-O2 \
-	-g3 \
-	-MD \
-	-MP \
-	-Wall \
-	-Wno-unused-function \
```
dumsum, line 87: Should we keep these warnings off or perhaps add them to toolchain snippet?

```	
-Wno-unused-function \
-Wno-bitwise-op-parentheses \
-Wno-shift-op-parentheses \ 
```

PR #209, opened 2025-09-19
Move to latest micropython

wom-bat:
Recent gcc and clang complain about micropython 1.24's source code This is fixed upstream, so use the latest upstream.

This is the minimum change set.

Tested on odroidc4 for the webserver and kitty examples.


wom-bat:
Tested also now on QEMU

Ivan-Velickovic:
I get the same error when building on Linux and macOS, so I don't think it's a macOS specific issue.

wom-bat:
I've now tested the firewall.  Some minor issues (mainly we were never using the USE_C_MODULES path correctly, and in fact do not need it.)

PR #210, opened 2025-09-22
fw_filter components update

cazb2:
Very similar to the old pull request into old_firewall.
Biggest differences between the prs, would be in meta.py,  mpfirewallport.c.

Has a build commit to get around micropython compiler error with most recent version of clang.


Courtney3141:
You'll have to fixup the failing CI checks (trailing whitespace and DCO which is signing off your commits with -s)

include/lions/firewall/filter.h
```diff
@@ -98,28 +103,88 @@ typedef struct fw_instance {
...
+ * Reserve an unused rule ID from the bitmap and mark it as allocated.
+ * Searches circularly starting from the last allocated ID position.
+ *
+ * @param bitmap pointer to the rule ID bitmap structure.
+ * @param rules_table_capacity total number of rule IDs available.
+ * @param res pointer to store the allocated rule ID.
+ *
+ * @return FILTER_ERR_OKAY if ID allocated successfully, FILTER_ERR_FULL if no IDs available.
+ */
+static fw_filter_err_t rules_reserve_id(fw_rule_id_bitmap_t *bitmap, uint16_t rules_table_capacity, uint16_t* res) {
```
Courtney3141: Since this function (and the following) both require a member(s) of the filter state, maybe it's worth just passing in the `fw_filter_state_t` as the first argument (also its the convention of the rest of the library).

Courtney3141: Then you could also add a quick one line check at the start that rule size < capacity, and if size == capacity return early.

include/lions/firewall/filter.h
```diff
@@ -162,17 +227,24 @@ typedef enum {
...
                                  fw_action_t default_action)
 {
-    state->rules = (fw_rule_t *)rules;
-    state->rules_capacity = rules_capacity;
-    state->internal_instances = (fw_instance_t *)internal_instances;
-    state->external_instances = (fw_instance_t *)external_instances;
+    state->rule_table = (fw_rule_table_t *)rules;
+    state->rules_capacity= rules_capacity;
+    state->rule_id_bitmap = (fw_rule_id_bitmap_t *) rule_id_bitmap;
+    uint16_t res;
```
Courtney3141: Rather than a print message here, I think an assertion is more appropriate as the system should likely crash if 0 is already allocated (this would cause issues in the future). Since there is no API for allocating a particular ID, maybe it's best to assert that 0 is not allocated first by accessing the data structure directly, then setting the 0th bit to 1 manually.

Courtney3141: Also, don't you need to increment rule size here to account for the first, default entry?

include/lions/firewall/filter.h
```diff
@@ -148,7 +213,17 @@ typedef enum {
     FILTER_RET_ERR = 0,
     FILTER_RET_RULE_ID = 1
 } fw_ret_args_t;
-
+static fw_filter_err_t fw_filter_add_rule(fw_filter_state_t *state,
```
Courtney3141: Should be a newline here

Courtney3141: Actually, rather than adding an additional declaration, just move the definition of `fw_filter_add_rule` before `fw_filter_state_init` - there is no real requirement that the init definition comes first.

include/lions/firewall/filter.h
```diff
@@ -162,18 +237,23 @@ typedef enum {
...
+    state->rule_table = (fw_rule_table_t *)rules;
+    state->rules_capacity= rules_capacity;
+    state->rule_id_bitmap = (fw_rule_id_bitmap_t *) rule_id_bitmap;
     state->instances_capacity = instances_capacity;
-    state->default_action = default_action;
+    state->internal_instances_table = (fw_instances_table_t *)internal_instances;
+    state->external_instances_table = (fw_instances_table_t *)external_instances;
+    uint16_t rule_id;
+    // add the the default rule
+    fw_filter_add_rule(state, 0, 0,0,0,0,0,true,true,default_action,&rule_id);
```
Courtney3141: The logic of the filter component is implicitly relying on two things here:
1. The `rule_id` of the default rule is 0
2. The index of the default rule in the rule table is 0 (hence why our `for` loops can start from 1)

If either of these are not true, the entire filter won't work.

You are able to check here that the `rule_id` of the default rule is 0, but you *don't* check that the index of the default rule in the table is 0, and the filter library doesn't really have an API to do this. Since we only need to do this once for the default rule, I don't think it is worth modifying the API.

I think in the special case of the default rule, you shouldn't use the API, and instead you should directly modify the data structures as follows:
1. Assert that the `rule_id` 0 is free inside the bitmap.
2. Allocate 0 (e.g. set the 0th bit in the bitmap manually).
3. Assert that the size of the rule table is 0 (implying that index 0 is free to be used).
4. Set the fields of the 0th entry manually.

If you do it this way, it guarantees that if someone who didn't quite understand the filter component somehow managed to add rules to the table before the `init` function was called, the filter would subsequently crash in a predictable way.

include/lions/firewall/filter.h
```diff
@@ -270,8 +342,13 @@ static fw_filter_err_t fw_filter_add_rule(fw_filter_state_t *state,
     empty_slot->src_port_any = src_port_any;
     empty_slot->dst_port_any = dst_port_any;
     empty_slot->action = action;
-    *rule_id = empty_slot - state->rules;
 
+    if (rules_reserve_id(state, rule_id) != FILTER_ERR_OKAY) {
+        return FILTER_ERR_FULL;
+    }
+    empty_slot->rule_id = *rule_id;
+    sddf_printf("allocating rule rule_id %d", *rule_id);
```
Courtney3141: Remove this print

include/lions/firewall/filter.h
```diff
@@ -270,8 +342,13 @@ static fw_filter_err_t fw_filter_add_rule(fw_filter_state_t *state,
     empty_slot->src_port_any = src_port_any;
     empty_slot->dst_port_any = dst_port_any;
     empty_slot->action = action;
-    *rule_id = empty_slot - state->rules;
 
+    if (rules_reserve_id(state, rule_id) != FILTER_ERR_OKAY) {
```
Courtney3141: Since we have already checked at the beginning of the function that the table is not full, this implies that the bitmap is also not full. If we fail to allocate a `rule_id` here that indicates that something has gone badly wrong (i.e. our bitmap and table are out of sync). Therefore we should instead _assert_ that the return value of `rules_reserve_id` is `FILTER_ERR_OKAY`

include/lions/firewall/filter.h
```diff
@@ -297,19 +374,14 @@ static fw_filter_err_t fw_filter_add_instance(fw_filter_state_t *state,
                                               bool default_action,
```
Courtney3141: The `default_action` boolean is no longer required now - check the changes you have made to see how it used to be used (we had a special case for default action in the filter code but now we don't, it's just indicated with `rule_id=0`). You need to remove this parameter and propagate the change to all the filters.

include/lions/firewall/filter.h
```diff
@@ -98,28 +103,88 @@ typedef struct fw_instance {
     uint16_t rule_id;
     /* instance was created after matching with the filter's default action */
     bool default_action;
```
Courtney3141: This is no longer needed as it is represented by `rule_id` being 0

include/lions/firewall/filter.h
```diff
@@ -445,21 +506,21 @@ static fw_filter_err_t fw_filter_remove_instances(fw_filter_state_t *state,
                                                   bool default_action,
```
Courtney3141: You need to rethink default action here, again it is no longer used.

include/lions/firewall/filter.h
```diff
@@ -445,21 +506,21 @@ static fw_filter_err_t fw_filter_remove_instances(fw_filter_state_t *state,
...
             continue;
         }
-
         if (!default_action && (rule_id != instance->rule_id)) {
+            i++;
             continue;
         }
 
-        instance->valid = false;
+        state->internal_instances_table->instances[i] = state->internal_instances_table->instances[state->internal_instances_table->size - 1];
```
Courtney3141, line 526: What happens if `i == size-1`...

include/lions/firewall/filter.h
```diff
@@ -502,8 +563,20 @@ static fw_filter_err_t fw_filter_update_default_action(fw_filter_state_t *state,
 static fw_filter_err_t fw_filter_remove_rule(fw_filter_state_t *state,
                                              uint16_t rule_id)
 {
-    fw_rule_t *rule = state->rules + rule_id;
-    if (rule_id >= state->rules_capacity || !rule->valid) {
+    if (rule_id == DEFAULT_RULE) {
```
Courtney3141: You should also check here that the corresponding bit is set in the bitmap

include/lions/firewall/filter.h
```diff
@@ -377,13 +442,9 @@ static fw_action_t fw_filter_find_action(fw_filter_state_t *state,
...
 
     /* Check rules */
-    fw_rule_t *match = NULL;
-    for (uint16_t i = 0; i < state->rules_capacity; i++) {
-        fw_rule_t *rule = state->rules + i;
-
-        if (!rule->valid) {
-            continue;
-        }
+    fw_rule_t *match = state->rule_table->rules;
```
Courtney3141: I think you should add a comment here making it clear that we can assume any traffic matches with the default rule which is stored in entry 0.

include/lions/firewall/filter.h
```diff
@@ -424,7 +485,7 @@ static fw_action_t fw_filter_find_action(fw_filter_state_t *state,
     }
 
     if (match) {
```
Courtney3141: This doesn't need to be in an if statement anymore as we always match with at least the default rule

include/lions/firewall/filter.h
```diff
@@ -424,7 +485,7 @@ static fw_action_t fw_filter_find_action(fw_filter_state_t *state,
...
-        *rule_id = match - state->rules;
+        *rule_id = match->rule_id;
         return match->action;
     }
```
Courtney3141: Also, this function can no longer return `FILTER_ACT_NONE`, so there is no longer any use for that enum - you should remove it.

include/lions/firewall/filter.h
```diff
@@ -98,28 +103,88 @@ typedef struct fw_instance {
...
+ * @param res pointer to store the allocated rule ID.
+ *
+ * @return FILTER_ERR_OKAY if ID allocated successfully, FILTER_ERR_FULL if no IDs available.
+ */
+static fw_filter_err_t rules_reserve_id(fw_filter_state_t* state, uint16_t* res) {
+    if (state->rule_table->size >= state->rules_capacity) {
+        return FILTER_ERR_FULL;
+    }
+
+    uint16_t total_ids = state->rules_capacity;
```
Courtney3141: No need to redefine a new variable here.

examples/firewall/filters/icmp_filter.c
```diff
@@ -49,18 +49,6 @@ static void filter(void)
             fw_action_t action = fw_filter_find_action(&filter_state, icmp_hdr->src_ip, ICMP_FILTER_DUMMY_PORT,
```
Courtney3141: `default_action` defined above is no longer needed, and always `false`. Same in all other filters.

examples/firewall/meta.py
```diff
@@ -132,11 +132,13 @@ def calculate_size(self):
 
 filter_rules_region = FirewallMemoryRegions("fw_rule_size",
                                          256,
-                                         lambda x: x.capacity * x.entry_size)
+                                         lambda x: x.entry_size + x.capacity * x.entry_size)
```
Courtney3141: This is not quite right, see `routing_table_region` for an example. It should be:

```
aligned_size(uint16_t size) + x.capacity * x.entry_size (routing table array size)
```

Since `fw_rule` contains `uint32_t`, the `uint16_t` size field in `fw_rule_table_t` will be padded to 4 bytes, so the lambda function should be:
`4 + x.capacity * x.entry_size`

examples/firewall/meta.py
```diff
@@ -132,11 +132,13 @@ def calculate_size(self):
...
 
 filter_instances_region = FirewallMemoryRegions("fw_instance_size",
                                          512,
-                                         lambda x: x.capacity * x.entry_size)
+                                         lambda x: x.entry_size + x.capacity * x.entry_size)
```
Courtney3141: This is also incorrect

examples/firewall/meta.py
```diff
@@ -132,11 +132,13 @@ def calculate_size(self):
...
+filter_rule_bitmap_capacity = (filter_rules_region.capacity + 63) // 64
```
Courtney3141: Could you create a FirewallMemoryRegions object for this? Since it doesn't require reading from the elf, you can use the `dma_buffer_region` as an example.

examples/firewall/meta.py
```diff
@@ -598,6 +600,11 @@ def generate(sdf_file: str, output_dir: str, dtb: DeviceTree):
             network["in_net"].add_client_with_copier(filter_pd, tx=False)
             network["configs"][in_virt].active_client_protocols.append(protocol)
 
+             # create bitmap
+            mr = MemoryRegion(sdf, "rules_id_bitmap" + "_" + filter_pd.name, filter_rule_bitmap_region_size);
```
Courtney3141: Since there are so many memory regions created in this file, I think it is best to use a more specific name e.g. possibly `rule_bitmap_mr`

components/micropython/modfirewall.c
```diff
@@ -477,42 +474,26 @@ static mp_obj_t rule_get_nth(mp_obj_t interface_idx_in, mp_obj_t protocol_in,
         mp_raise_OSError(OS_ERR_INVALID_PROTOCOL);
         return mp_const_none;
     }
-
-    if (rule_idx >= webserver_state[interface_idx].num_rules[protocol_match] ||
+    if (rule_idx >= webserver_state[interface_idx].filter_states[protocol_match].rule_table->size ||
```
Courtney3141: Re-add deleted newline. Also second or condition is no longer required, since `size <= capacity`

components/micropython/modfirewall.c
```diff
@@ -477,42 +474,26 @@ static mp_obj_t rule_get_nth(mp_obj_t interface_idx_in, mp_obj_t protocol_in,
         mp_raise_OSError(OS_ERR_INVALID_PROTOCOL);
```
Courtney3141, line 473: I think it may be worth returning an error in this function if `rule_idx=0`, since if we have a specific function for getting the default rule `filter_get_default_action`, we should probably ensure that it is always used.

include/lions/firewall/filter.h
Courtney3141: Never mind, it is not a problem since we decrement the size, whoops! looks good.

include/lions/firewall/filter.h
```diff
@@ -35,10 +40,8 @@ static const char *fw_filter_err_str[] = {
 };
 
 typedef enum {
-    /* no rule exists */
-    FILTER_ACT_NONE,
     /* allow traffic */
-	FILTER_ACT_ALLOW,
+	FILTER_ACT_ALLOW = 1,
```
Courtney3141: No need

include/lions/firewall/filter.h
```diff
@@ -96,30 +99,87 @@ typedef struct fw_instance {
...
+/**
+ * Reserve an unused rule ID from the bitmap and mark it as allocated.
+ * Searches circularly starting from the last allocated ID position.
+ *
+ * @param state pointer to the filter state.
+ * @param res pointer to store the allocated rule ID.
+ *
+ * @return FILTER_ERR_OKAY if ID allocated successfully, FILTER_ERR_FULL if no IDs available.
+ */
+static fw_filter_err_t rules_reserve_id(fw_filter_state_t* state, uint16_t* res) {
```
Courtney3141: newline, assert success

Courtney3141: add internal error

include/lions/firewall/filter.h
```diff
@@ -162,18 +222,35 @@ typedef enum {
...
-    state->default_action = default_action;
+    state->internal_instances_table = (fw_instances_table_t *)internal_instances;
+    state->external_instances_table = (fw_instances_table_t *)external_instances;
+
+    uint64_t mask = 1ULL;
+    assert((state->rule_id_bitmap->id_bitmap[DEFAULT_RULE] & mask) == 0);
+    state->rule_id_bitmap->id_bitmap[DEFAULT_RULE] |= mask;
+    state->rule_id_bitmap->last_allocated_rule_id = DEFAULT_RULE;
+
+    assert(state->rule_table->size == 0);
```
Courtney3141: comment

Courtney3141: default rule id and calculation

include/lions/firewall/filter.h
```diff
@@ -96,30 +99,87 @@ typedef struct fw_instance {
...
+    return FILTER_ERR_FULL;
+}
+/**
+ * Free a previously allocated rule ID by clearing its bit in the bitmap.
+ * The default rule (ID 0) cannot be freed and attempts to free it are ignored.
+ *
+ * @param state pointer to the filter state
+ * @param id rule ID to free.
+ */
+static void rules_free_id(fw_filter_state_t* state, uint16_t id) {
```
Courtney3141: check ID in capacity, return error

include/lions/firewall/filter.h
```diff
@@ -376,14 +434,10 @@ static fw_action_t fw_filter_find_action(fw_filter_state_t *state,
...
-    fw_rule_t *match = NULL;
-    for (uint16_t i = 0; i < state->rules_capacity; i++) {
-        fw_rule_t *rule = state->rules + i;
-
-        if (!rule->valid) {
-            continue;
-        }
+    /* Check rules for best match otherwise we match with the default rule */
+    fw_rule_t *match = &state->rule_table->rules[DEFAULT_RULE];
+    for (uint16_t i = 1; i < state->rule_table->size; i++) {
```
Courtney3141: default +1

include/lions/firewall/filter.h
```diff
@@ -476,17 +520,17 @@ static fw_filter_err_t fw_filter_remove_instances(fw_filter_state_t *state,
 static fw_filter_err_t fw_filter_update_default_action(fw_filter_state_t *state,
                                                        fw_action_t new_action)
 {
-    fw_action_t old_action = state->default_action;
+    fw_action_t old_action = state->rule_table->rules->action;
```
Courtney3141: use macro

Courtney3141 (approved):

PR #221, opened 2025-10-10
Move fs helpers to library

dumsum:
I've been looking at this as part of wasm fs stuff but this kind of stands alone so figured I'd open for feedback.

There was very little micropython specific functionality in `fs_helpers`. The main change in this commit (aside from moving the files) is to abstract these specifics away by requiring the provision of a custom "wait" function for `fs_command_blocking` (set with the new function `fs_set_blocking_wait`, on init). Includes for `fs_helpers.h` are also updated.

`blocking_wait` would in general simply be `microkit_cothread_wait_on_channel`, but MP does some extra interrupt handling as well hence the need for a custom function, which also helps avoid an explicit dependency on `libmicrokitco`.

### DIFF
--- a/components/micropython/Makefile
+++ b/components/micropython/Makefile
@@ -70,7 +70,7 @@ SRC_C = \
 	modfs_raw.c \
 	vfs_fs_file.c \
 	vfs_fs.c \
-	fs_helpers.c \
+	fs/helpers.c \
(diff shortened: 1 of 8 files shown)


dumsum:
Included in https://github.com/au-ts/lionsos/pull/236 instead

PR #231, opened 2025-10-17
Add docker/qemu support for the firewall

Courtney3141:
This PR primarily adds docker/qemu support to the firewall, providing a docker file to build a container with the network infrastructure needed to test the firewall. The firewall itself can then be run inside the container using qemu.

Additionally, this PR also:
- allows the firewall to calculate checksums when required
- fixes some inconsistencies with storing port numbers little-endian
- updates the definitions of Ethernet and IP protocol headers
- updates the firewall network virt rx - client interface, allowing both an eth-type and sub-type (typical ARP opcode or IP protocol) to be specified

### DIFF
--- a/examples/firewall/docker/README.md
+++ b/examples/firewall/docker/README.md
@@ -13,4 +13,3 @@ container.
 
 Further building and testing information can be found on the lionsos
 [website](https://https://lionsos.org/docs/examples/firewall/).
-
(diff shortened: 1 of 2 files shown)


examples/firewall/docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# Copyright 2025, UNSW
+# SPDX-License-Identifier: BSD-2-Clause
+
+FROM ubuntu:latest
```
Ivan-Velickovic, line 4: You probably want to pin it so it doesn't break underneath you when a new version comes out.

Ivan-Velickovic: For example, `FROM ubuntu:24.04`

Courtney3141: Yes good idea, I'll do that when I next get the chance

PR #236, opened 2025-10-26
POSIX File I/O

dumsum:
Implements rudimentary file io in the POSIX layer.
Cleans up POSIX syscall implementation source files.
Adds a basic example demonstrating the use of standard POSIX APIs for file io.


Ivan-Velickovic (approved):
More sanitisation/error handling to do for the POSIX functionality but this is a big improvement right now.

Probably a bit later once this matures, we should add the POSIX example to lionsos.org.

PR #242, opened 2025-11-03
dwarf for firewall components

cazb2:
using llvm-dwarf dump to obtain the size of the structs used for memory memory regions in various fw components

### DIFF
--- a/examples/firewall/meta.py
+++ b/examples/firewall/meta.py
@@ -10,7 +10,8 @@ from ctypes import *
 from importlib.metadata import version
 import ipaddress
 from board import BOARDS
-assert version('sdfgen').split(".")[1] == "27", "Unexpected sdfgen version"
+
+assert version("sdfgen").split(".")[1] == "27", "Unexpected sdfgen version"
 
 from sdfgen_helper import *
 
@@ -23,14 +24,15 @@ Channel = SystemDescription.Channel
 ext_net = 0
 int_net = 1
 
-macs = [[0x00, 0x01, 0xc0, 0x39, 0xd5, 0x18], # External network
-        [0x00, 0x01, 0xc0, 0x39, 0xd5, 0x10]] # Internal network
+macs = [
+    [0x00, 0x01, 0xC0, 0x39, 0xD5, 0x18],  # External network
+    [0x00, 0x01, 0xC0, 0x39, 0xD5, 0x10],
+]  # Internal network
 
-subnet_bits = [12, # External network
-               24] # Internal network
+subnet_bits = [12, 24]  # External network  # Internal network
 
-ips = ["172.16.2.1",  # External network
-       "192.168.1.1"] # Internal network
+ips = ["172.16.2.1", "192.168.1.1"]  # External network  # Internal network
+
 
 @dataclass
 class Board:
@@ -42,6 +44,7 @@ class Board:
     ethernet0: str
     ethernet1: str
 
+
 BOARDS: List[Board] = [
     Board(
         name="qemu_virt_aarch64",
@@ -50,7 +53,7 @@ BOARDS: List[Board] = [
         serial="pl011@9000000",
         timer="timer",
         ethernet0="virtio_mmio@a003c00",
-        ethernet1="virtio_mmio@a003e00"
+        ethernet1="virtio_mmio@a003e00",
     ),
     Board(
         name="imx8mp_iotgate",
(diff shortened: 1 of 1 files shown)


examples/firewall/meta.py
```diff
@@ -110,47 +110,52 @@ def calculate_size(self):
             sys.exit()
 
 # Firewall memory region object declarations, update region capacities here
-dma_buffer_queue_region = FirewallMemoryRegions("fw_buffer_queue_entry_size",
+dma_buffer_queue_region = FirewallMemoryRegions("routing.elf","net_buff_desc",
                                                 512,
                                                 lambda x: 16 + x.capacity * x.entry_size)
 
-dma_buffer_region = FirewallMemoryRegions(None,
+dma_buffer_region = FirewallMemoryRegions(None,None,
```
Courtney3141: Space between arguments

examples/firewall/meta.py
```diff
@@ -776,46 +781,25 @@ def generate(sdf_file: str, output_dir: str, dtb: DeviceTree):
...
+                    for i in range(len(dwarfdump)):
+                        if dwarfdump[i] == "DW_TAG_structure_type":
+                            if dwarfdump[i + 2] == f"(\"{mem_region.c_name}\")":
+                                assert dwarfdump[i + 3] == "DW_AT_byte_size"
+                                size_fmt = dwarfdump[i + 4].strip('(').strip(')')
+                                size = int(size_fmt, base=16)
+                                mem_region.entry_size = size
+                except:
+                    raise Exception(f"Error running the llvm-dwarf dump on {elf_file_name})")
         mem_region.calculate_size()
```
Courtney3141: Since we allow the case of region and entry size being 0, and elf file name not being defined (which typically shouldn't happen and likely means your region size will be 0), could you add an assertion that the calculated region size is not 0?

examples/firewall/meta.py
```diff
@@ -110,47 +110,52 @@ def calculate_size(self):
...
-dma_buffer_region = FirewallMemoryRegions(None,
+dma_buffer_region = FirewallMemoryRegions(None,None,
                                           dma_buffer_queue_region.capacity,
                                           lambda x: x.capacity * x.entry_size,
                                           2048)
 
-arp_queue_region = FirewallMemoryRegions("fw_arp_queue_entry_size",
+arp_queue_region = FirewallMemoryRegions("arp_requester.elf","fw_arp_request",
                                          512,
                                          lambda x: 16 + x.capacity * x.entry_size)
```
Courtney3141: I like what you have done with extracting the wrapper size as well! In this case I think you may as well do a similar thing and extract the size of the queue indeces as well e.g. `fw_queue_indeces` in include/lions/firewall/queue.h

examples/firewall/routing/routing.c
```diff
@@ -468,15 +468,4 @@ void notified(microkit_channel ch)
         tx_net = false;
         microkit_notify(router_config.tx_active.ch);
     }
```
Courtney3141, line 469: Good job with removing these! When I created these variables, I also added a few unnecessary header includes to the routing component as well. Could you see if you could remove those?

Courtney3141 (dismissed):
Nice job! Could you also please make a PR to update the corresponding docs here: https://lionsos.org/docs/examples/firewall/building/

The source repo is here: https://github.com/au-ts/lionsos.org

examples/firewall/meta.py
cazb2: when I was updating this, I thought that it made sense to use the struct fw_queue as the structure to get the entry size from despite this slightly overestimating the required size

examples/firewall/routing/routing.c
cazb2: could only find two includes in routing.c that were added 2 weeks ago that were safe to remove

examples/firewall/meta.py
```diff
@@ -73,84 +74,108 @@ def round_up_to_Page(region_size: int) -> int:
...
+            self.c_name = c_name
+            self.entry_size = 0
+            self.size_formula_bytes = size_formula_bytes
+            self.size = 0
+        else:
+            raise Exception("Invalid Argument Combination")
+
+    def calculate_size(self):
+        if self.size != 0:
+            print("Size has already been calculated, this is probably an error")
```
Courtney3141: more helpful error messages

examples/firewall/meta.py
```diff
@@ -73,84 +74,108 @@ def round_up_to_Page(region_size: int) -> int:
...
-        self.capacity = capacity
-        # Formula for calculating memory region size from class object
+    def __init__(
+        self, *, unaligned_size = None, dependent_type_info: List[DataStructureTypeInfo]|None = None, region_size_formula = lambda list: sum(item.get_size() for item in list)
+    ):
+        if unaligned_size:
+            self.region_size = round_up_to_Page(unaligned_size)
+        elif dependent_type_info:
+            self.region_size = 0
+            self.dependent_type_info = dependent_type_info
```
Courtney3141: better parameter names

PR #244, opened 2025-11-05
Make micrpython-submodules if necessary

wom-bat:
On a fresh checkout of LionsOS, you'll get the message when bulding anything to do with micropython,
  'Error: micropython-lib submodule is not initialized. Run 'make
  submodules''

Unfortunately it's not so easy to work out where or how to do this. So fix the issue here in the Makefile.


Ivan-Velickovic:
That will update every MicroPython submodule right? I want to avoid that since there are many of them, it would be better if we just update `lib/micropython-lib`.

wom-bat:
I've changed from an implicit dependency to an 
explicit MAKE command, in the hopes that will fix it on MAC.

I've also rebased on tip of main, as there were I2C changes.

Issue #194, opened 2025-09-03
Improve the ICMP module component


Courtney3141:
For an overview of different types of ICMP messages see:

https://www.computernetworkingnotes.com/networking-tutorials/icmp-error-messages-and-format-explained.html.

PR #253, opened 2026-01-06
Add ICMP extensions

DaiDaiDai123:
Issue #194 

1. Added ICMP ping support for the firewall by creating `enqueue_icmp_ping` in `routing.c`. The webservre can toggle the supprt for each interface.
2. Add support for UDP Reject. An ICMP `Destination Unreachable` is sent when rejected.

### DIFF
--- a/components/micropython/modfirewall.c
+++ b/components/micropython/modfirewall.c
@@ -10,6 +10,7 @@
 #include <sddf/util/printf.h>
 #include <lions/firewall/config.h>
 #include <lions/firewall/filter.h>
+#include <lions/firewall/ip.h>
 #include <lions/firewall/routing.h>
 
 #include "mpfirewallport.h"
@@ -28,7 +29,8 @@ typedef enum {
     OS_ERR_INVALID_ROUTE_NUM, /* Invalid route number supplied to route_get_nth */
     OS_ERR_INVALID_RULE_NUM,  /* Invalid route number supplied to rule_get_nth */
     OS_ERR_OUT_OF_MEMORY,     /* Data structures full */
-    OS_ERR_INTERNAL_ERROR     /* Unknown internal error */
+    OS_ERR_INTERNAL_ERROR,    /* Unknown internal error */
+    OS_ERR_UNSUPPORTED_ACTION /* Unsupported action for selected protocol */
 } fw_os_err_t;
 
 static const char *fw_os_err_str[] = {
@@ -44,8 +46,22 @@ static const char *fw_os_err_str[] = {
     "Route number supplied is greater than the number of routes.",
     "Rule number supplied is the default action rule index, or greater than the number of rules.",
     "Internal data structures are already at capacity.",
-    "Unknown internal error."
+    "Unknown internal error.",
+    "Unsupported action for the protocol selected."
 };
+
+static bool is_action_supported_for_filter(fw_webserver_filter_config_t *filter, uint8_t action)
+{
+    if (action == 0 || action > FW_FILTER_NUM_ACTIONS) {
+        return false;
+    }
+
+    if (filter->actions[action - 1]) {
+        return true;
+    }
+
+    return false;
+}
 
 /* Convert a routing error to OS error */
 static fw_os_err_t fw_routing_err_to_os_err(fw_routing_err_t routing_err)
@@ -82,6 +98,8 @@ static fw_os_err_t filter_err_to_os_err(fw_filter_err_t filter_err)
         return OS_ERR_CLASH;
     case FILTER_ERR_INVALID_RULE_ID:
         return OS_ERR_INVALID_RULE_ID;
+    case FILTER_ERR_UNSUPPORTED_ACTION:
+        return OS_ERR_UNSUPPORTED_ACTION;
     default:
         return OS_ERR_INTERNAL_ERROR;
     }
@@ -187,6 +205,53 @@ static mp_obj_t route_delete(mp_obj_t interface_idx_in, mp_obj_t route_id_in)
 
 static MP_DEFINE_CONST_FUN_OBJ_2(route_delete_obj, route_delete);
 
+/* Enable or disable ICMP ping responses on an interface */
+static mp_obj_t ping_response_set(mp_obj_t interface_idx_in, mp_obj_t enable_in) {
+    uint8_t interface_idx = mp_obj_get_int(interface_idx_in);
+    if (interface_idx >= FW_NUM_INTERFACES) {
+        sddf_dprintf("WEBSERVER|LOG: %s\n",
+                    fw_os_err_str[OS_ERR_INVALID_INTERFACE]);
+        mp_raise_OSError(OS_ERR_INVALID_INTERFACE);
+        return mp_const_none;
+    }
+
+    bool enable = mp_obj_is_true(enable_in);
+
+    microkit_mr_set(0, enable ? 1 : 0);
+    microkit_msginfo msginfo =
+        microkit_ppcall(fw_config.interfaces[interface_idx].router.routing_ch,
+                    microkit_msginfo_new(FW_SET_PING_RESPONSE, 1));
+
+    uint32_t result = microkit_mr_get(0);
+    if (result != 0) {
+        sddf_dprintf("WEBSERVER|LOG: Failed to set ping response\n");
+        mp_raise_OSError(OS_ERR_INTERNAL_ERROR);
+        return mp_const_none;
+    }
+
+    /* Store the state of the ping response */
+    webserver_state[interface_idx].ping_enabled = enable;
+
+    return mp_const_none;
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_2(ping_response_set_obj, ping_response_set);
+
+/* Get status of ICMP ping responses on an interface */
+static mp_obj_t ping_response_get(mp_obj_t interface_idx_in) {
+    uint8_t interface_idx = mp_obj_get_int(interface_idx_in);
+    if (interface_idx >= FW_NUM_INTERFACES) {
+        sddf_dprintf("WEBSERVER|LOG: %s\n",
+                    fw_os_err_str[OS_ERR_INVALID_INTERFACE]);
+        mp_raise_OSError(OS_ERR_INVALID_INTERFACE);
+        return mp_const_none;
+    }
+
+    return mp_obj_new_bool(webserver_state[interface_idx].ping_enabled);
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_1(ping_response_get_obj, ping_response_get);
+
 /* Count the number of routes in an interface routing table */
 static mp_obj_t route_count(mp_obj_t interface_idx_in)
 {
@@ -273,6 +338,13 @@ static mp_obj_t rule_add(mp_uint_t n_args, const mp_obj_t *args)
     if (protocol_match == fw_config.interfaces[interface_idx].num_filters) {
         sddf_dprintf("WEBSERVER|LOG: %s\n", fw_os_err_str[OS_ERR_INVALID_PROTOCOL]);
         mp_raise_OSError(OS_ERR_INVALID_PROTOCOL);
+        return mp_const_none;
+    }
+
+    if (!is_action_supported_for_filter(&fw_config.interfaces[interface_idx].filters[protocol_match],
+                                        action)) {
+        sddf_dprintf("WEBSERVER|LOG: %s\n", fw_os_err_str[OS_ERR_UNSUPPORTED_ACTION]);
+        mp_raise_OSError(OS_ERR_UNSUPPORTED_ACTION);
         return mp_const_none;
     }
 
@@ -392,6 +464,13 @@ static mp_obj_t filter_set_default_action(mp_obj_t interface_idx_in, mp_obj_t pr
         return mp_const_none;
     }
 
+    if (!is_action_supported_for_filter(&fw_config.interfaces[interface_idx].filters[protocol_match],
+                                        action)) {
+        sddf_dprintf("WEBSERVER|LOG: %s\n", fw_os_err_str[OS_ERR_UNSUPPORTED_ACTION]);
+        mp_raise_OSError(OS_ERR_UNSUPPORTED_ACTION);
+        return mp_const_none;
+    }
+
     microkit_mr_set(FILTER_ARG_ACTION, action);
     microkit_msginfo msginfo = microkit_ppcall(fw_config.interfaces[interface_idx].filters[protocol_match].ch,
                                                microkit_msginfo_new(FW_SET_DEFAULT_ACTION, 1));
@@ -484,19 +563,21 @@ static mp_obj_t rule_get_nth(mp_obj_t interface_idx_in, mp_obj_t protocol_in, mp
 static MP_DEFINE_CONST_FUN_OBJ_3(rule_get_nth_obj, rule_get_nth);
 
 static const mp_rom_map_elem_t lions_firewall_module_globals_table[] = {
-    { MP_OBJ_NEW_QSTR(MP_QSTR___name__), MP_ROM_QSTR(MP_QSTR_lions_firewall) },
-    { MP_ROM_QSTR(MP_QSTR_interface_mac_get), MP_ROM_PTR(&interface_get_mac_obj) },
-    { MP_ROM_QSTR(MP_QSTR_interface_ip_get), MP_ROM_PTR(&interface_get_ip_obj) },
-    { MP_ROM_QSTR(MP_QSTR_route_add), MP_ROM_PTR(&route_add_obj) },
-    { MP_ROM_QSTR(MP_QSTR_route_delete), MP_ROM_PTR(&route_delete_obj) },
-    { MP_ROM_QSTR(MP_QSTR_route_count), MP_ROM_PTR(&route_count_obj) },
-    { MP_ROM_QSTR(MP_QSTR_route_get_nth), MP_ROM_PTR(&route_get_nth_obj) },
-    { MP_ROM_QSTR(MP_QSTR_rule_add), MP_ROM_PTR(&rule_add_obj) },
-    { MP_ROM_QSTR(MP_QSTR_rule_delete), MP_ROM_PTR(&rule_delete_obj) },
-    { MP_ROM_QSTR(MP_QSTR_rule_count), MP_ROM_PTR(&rule_count_obj) },
-    { MP_ROM_QSTR(MP_QSTR_rule_get_nth), MP_ROM_PTR(&rule_get_nth_obj) },
-    { MP_ROM_QSTR(MP_QSTR_filter_get_default_action), MP_ROM_PTR(&filter_get_default_action_obj) },
-    { MP_ROM_QSTR(MP_QSTR_filter_set_default_action), MP_ROM_PTR(&filter_set_default_action_obj) },
+    { MP_OBJ_NEW_QSTR(MP_QSTR___name__), MP_ROM_QSTR(MP_QSTR_lions_firewall)},
+    { MP_ROM_QSTR(MP_QSTR_interface_mac_get), MP_ROM_PTR(&interface_get_mac_obj)},
+    { MP_ROM_QSTR(MP_QSTR_interface_ip_get), MP_ROM_PTR(&interface_get_ip_obj)},
+    { MP_ROM_QSTR(MP_QSTR_route_add), MP_ROM_PTR(&route_add_obj)},
+    { MP_ROM_QSTR(MP_QSTR_route_delete), MP_ROM_PTR(&route_delete_obj)},
+    { MP_ROM_QSTR(MP_QSTR_route_count), MP_ROM_PTR(&route_count_obj)},
+    { MP_ROM_QSTR(MP_QSTR_route_get_nth), MP_ROM_PTR(&route_get_nth_obj)},
+    { MP_ROM_QSTR(MP_QSTR_ping_response_set), MP_ROM_PTR(&ping_response_set_obj)},
+    { MP_ROM_QSTR(MP_QSTR_ping_response_get), MP_ROM_PTR(&ping_response_get_obj)},
+    { MP_ROM_QSTR(MP_QSTR_rule_add), MP_ROM_PTR(&rule_add_obj)},
+    { MP_ROM_QSTR(MP_QSTR_rule_delete), MP_ROM_PTR(&rule_delete_obj)},
+    { MP_ROM_QSTR(MP_QSTR_rule_count), MP_ROM_PTR(&rule_count_obj)},
+    { MP_ROM_QSTR(MP_QSTR_rule_get_nth), MP_ROM_PTR(&rule_get_nth_obj)},
+    { MP_ROM_QSTR(MP_QSTR_filter_get_default_action), MP_ROM_PTR(&filter_get_default_action_obj)},
+    { MP_ROM_QSTR(MP_QSTR_filter_set_default_action), MP_ROM_PTR(&filter_set_default_action_obj)},
 };
 
 static MP_DEFINE_CONST_DICT(lions_firewall_module_globals, lions_firewall_module_globals_table);
--- a/components/micropython/mpfirewallport.h
+++ b/components/micropython/mpfirewallport.h
@@ -27,6 +27,7 @@ typedef struct fw_webserver_interface_state {
     fw_routing_table_t *routing_table;
     fw_filter_state_t filter_states[FW_MAX_FILTERS];
     uint16_t num_rules[FW_MAX_FILTERS];
+    bool ping_enabled;
 } fw_webserver_interface_state_t;
 
 extern fw_webserver_interface_state_t webserver_state[FW_NUM_INTERFACES];
--- a/examples/firewall/docker/scripts/net_setup.sh
+++ b/examples/firewall/docker/scripts/net_setup.sh
@@ -1,8 +1,6 @@
 # Copyright 2025, UNSW
 # SPDX-License-Identifier: BSD-2-Clause
 
-#!/bin/bash
-
 # -- Container script for setting up the network -- #
 
 # Create bridges to connect namespaces to taps
--- a/examples/firewall/filters/icmp_filter.c
+++ b/examples/firewall/filters/icmp_filter.c
@@ -24,11 +24,23 @@ __attribute__((__section__(".net_client_config"))) net_client_config_t net_confi
 net_queue_handle_t rx_queue;
 net_queue_handle_t tx_queue;
 fw_queue_t router_queue;
+fw_queue_t icmp_queue;
 
 /* Holds filtering rules and state */
 fw_filter_state_t filter_state;
 
 #define ICMP_FILTER_DUMMY_PORT 0
+
+/* ICMP request queue to send unreachable messages to ICMP module */
+static bool notify_icmp;
+
+static bool enqueue_icmp_unreachable(net_buff_desc_t buffer)
+{
+    uintptr_t pkt_vaddr = (uintptr_t)(net_config.rx_data.vaddr + buffer.io_or_offset);
+    bool enqueued = icmp_enqueue_error(&icmp_queue, ICMP_DEST_UNREACHABLE, ICMP_DEST_PORT_UNREACHABLE, pkt_vaddr);
+    notify_icmp |= enqueued;
+    return enqueued;
+}
 
 static void filter(void)
 {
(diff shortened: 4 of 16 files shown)


components/micropython/modfirewall.c
```diff
@@ -190,6 +190,35 @@ static mp_obj_t route_delete(mp_obj_t interface_idx_in, mp_obj_t route_id_in) {
...
+    if (interface_idx >= FW_NUM_INTERFACES) {
+        sddf_dprintf("WEBSERVER|LOG: %s\n",
+                    fw_os_err_str[OS_ERR_INVALID_INTERFACE]);
+        mp_raise_OSError(OS_ERR_INVALID_INTERFACE);
+        return mp_const_none;
+    }
+
+    bool enable = mp_obj_is_true(enable_in);
+
+    seL4_SetMR(0, enable ? 1 : 0);
```
Kswin01: If you could switch these `seL4_SetMR` and `seL4_GetMR` occurances with `microkit_mr_set` and `microkit_mr_get` that'd be great

examples/firewall/docker/scripts/firewall_configuration.sh
```diff
@@ -56,4 +56,4 @@ export INT_HOST_IP=192.168.1.100
 export EXT_BAD_HOST_IP=172.16.2.201
 
 # Internal non-existant host IP address
-export INT_BAD_HOST_IP=192.168.1.101
+export INT_BAD_HOST_IP=192.168.1.101
```
Kswin01: If you could add whitespace back here to remove this diff

examples/firewall/filters/udp_filter.c
```diff
@@ -25,10 +26,37 @@ __attribute__((__section__(".net_client_config"))) net_client_config_t net_confi
...
 fw_queue_t router_queue;
+fw_queue_t icmp_queue;
 
 /* Holds filtering rules and state */
 fw_filter_state_t filter_state;
 
+/* ICMP request queue to send unreachable messages to ICMP module */
+static bool notify_icmp;
+
+static int enqueue_icmp_unreachable(net_buff_desc_t buffer)
```
Kswin01: This seems to be the same as the function in `routing.c`. Might be better to add these ICMP functions into a seperate header file and pass through the req code and type?

examples/firewall/filters/udp_filter.c
```diff
@@ -89,6 +117,7 @@ static void filter(void)
                             fw_frmt_str[filter_config.interface], rule_id,
                             ipaddr_to_string(ip_hdr->src_ip, ip_addr_buf0), htons(udp_hdr->src_port),
                             ipaddr_to_string(ip_hdr->dst_ip, ip_addr_buf1), htons(udp_hdr->dst_port));
+
```
Kswin01: whitespace
```suggestion
```

Kswin01: Sorry for these pedantic whitespace comments, but they cause the style checker to fail

examples/firewall/filters/udp_filter.c
```diff
@@ -208,6 +262,11 @@ void init(void)
     fw_queue_init(&router_queue, filter_config.router.queue.vaddr,
         sizeof(net_buff_desc_t),  filter_config.router.capacity);
 
+    if (filter_config.icmp_module.queue.vaddr) {
```
Kswin01: In what case do we not connect the udp filter and the icmp module? Is this check needed?

examples/firewall/ui_server.py
```diff
@@ -52,8 +52,8 @@
 numInterfaces = 2
 
 interfaceStringsRouters = [
-    "internal",
     "external",
+    "internal"
```
Kswin01: Is this supposed to have been changed?

examples/firewall/routing/routing.c
```diff
@@ -88,6 +96,69 @@ static int enqueue_icmp_unreachable(net_buff_desc_t buffer)
     return err;
 }
 
+static void enqueue_icmp_ping(net_buff_desc_t buffer)
```
Kswin01: same as in udp filter, can we move these to a seperate file?

examples/firewall/routing/routing.c
```diff
@@ -420,6 +518,16 @@ seL4_MessageInfo_t protected(microkit_channel ch, microkit_msginfo msginfo)
         seL4_SetMR(ROUTER_RET_ERR, err);
         return microkit_msginfo_new(0, 1);
     }
+    case FW_SET_PING_RESPONSE: {
+        ping_response_enabled = (bool)seL4_GetMR(0);
```
Kswin01:
```suggestion
        ping_response_enabled = (bool)microkit_mr_get(0);
```

examples/firewall/routing/routing.c
```diff
@@ -420,6 +518,16 @@ seL4_MessageInfo_t protected(microkit_channel ch, microkit_msginfo msginfo)
...
+        if (FW_DEBUG_OUTPUT) {
+            sddf_printf("%sRouter ping response %s\n",
+                fw_frmt_str[router_config.interface],
+                ping_response_enabled ? "enabled" : "disabled");
+        }
+        seL4_SetMR(0, 0); /* success */
```
Kswin01:
```suggestion
        microkit_mr_set(0, 0); /* success */
```

include/lions/firewall/routing.h
```diff
@@ -499,3 +500,26 @@ static void fw_routing_table_init(fw_routing_table_t **table,
...
+
+/**
+ * Check if the given IP address has a valid route in the routing table.
+ *
+ * @param table address of routing table.
+ * @param ip IP address to check for a valid route.
+ *
+ * @return true if a valid route exists, false otherwise.
+ */
+static bool fw_routing_table_is_last_hop(fw_routing_table_t *table,
```
Kswin01: This doesnt seem to be checking if last hop but rather if theres a valid route?

include/lions/firewall/icmp.h
```diff
@@ -78,18 +89,125 @@ typedef struct __attribute__((__packed__)) icmp_dest {
...
+
+/* ----------------- 8 - Echo Request / 0 - Echo Reply ---------------------------*/
+
+/* ICMP echo request/reply header fields (following common header) */
+typedef struct __attribute__((__packed__)) icmp_echo {
+    /* Identifier to match requests with replies */
+    uint16_t id;
+    /* Sequence number */
+    uint16_t seq;
+    /* Payload data follows */
```
Courtney3141: Is there a reason you don't define an array here, similar to how we do for `destination unreachable?`

components/micropython/modfirewall.c
```diff
@@ -190,6 +190,35 @@ static mp_obj_t route_delete(mp_obj_t interface_idx_in, mp_obj_t route_id_in) {
 
 static MP_DEFINE_CONST_FUN_OBJ_2(route_delete_obj, route_delete);
 
+/* Enable or disable ICMP ping responses on an interface */
```
Courtney3141, line 208: Shouldn't there be a similar function for _getting_ whether `ping` is enabled for each interface?

include/lions/firewall/filter.h
```diff
@@ -49,15 +49,18 @@ typedef enum {
     /* allow traffic, and additionally any return traffic */
     FILTER_ACT_CONNECT = 3,
     /* traffic is return traffic from a connect rule */
-    FILTER_ACT_ESTABLISHED
+    FILTER_ACT_ESTABLISHED = 4,
+    /* reject traffic for UDP traffic */
+    FILTER_UDP_REJECT
```
Courtney3141: Why can't we use the pre-existing `FILTER_ACT_DROP` here? And just have different behaviour within the UDP filter when traffic is dropped? Since both `FILTER_ACT_DROP` and `FILTER_UDP_REJECT` describe the same behaviour, it's just that in the case of UDP traffic the ICMP module sends a destination unreachable.

examples/firewall/filters/udp_filter.c
```diff
@@ -103,6 +121,22 @@ static void filter(void)
                         ipaddr_to_string(ip_hdr->src_ip, ip_addr_buf0), htons(udp_hdr->src_port),
                         ipaddr_to_string(ip_hdr->dst_ip, ip_addr_buf1), htons(udp_hdr->dst_port));
                 }
+            } else if (action == FILTER_UDP_REJECT) {
```
Courtney3141: See my comment in `filter.h` - unsure of why `FILTER_ACT_DROP` can't still be used.

examples/firewall/filters/udp_filter.c
```diff
@@ -116,6 +150,8 @@ static void filter(void)
     }
 
     if (returned) {
+        sddf_printf("%sUDP filter returned packets to RX virtualiser\n The action is %s\n",
```
Courtney3141: Needs to be wrapped in a `FW_DEBUG_OUTPUT` conditional (since this is normal behaviour).

examples/firewall/filters/udp_filter.c
```diff
@@ -196,6 +232,13 @@ void notified(microkit_channel ch)
         sddf_dprintf("%sUDP FILTER LOG: Received notification on unknown channel: %d!\n",
             fw_frmt_str[filter_config.interface], ch);
     }
+
+    if (notify_icmp) {
+        sddf_printf("%sUDP filter notifying ICMP module on channel %u\n",
```
Courtney3141: Likewise here - `FW_DEBUG_OUTPUT`

examples/firewall/ui_server.py
```diff
@@ -701,6 +714,7 @@ def rules(request, protocol):
           <option value="1">Allow</option>
           <option value="2">Drop</option>
           <option value="3">Connect</option>
+          <option value="5">Reject</option>
```
Courtney3141: I am thinking this should be the same as drop as mentioned.

examples/firewall/ui_server.py
```diff
@@ -988,5 +1004,68 @@ def css(request):
...
+    <h1>Ping Response Settings</h1>
+    <nav>
+      <a href="/">Home</a> | <a href="/routing_config">Routing Config</a> | <a href="/rules">Rules</a> | <a href="/interface">Interface</a> | <a href="/ping_settings">Ping Settings</a>
+    </nav>
+
+    <h2>Toggle Ping Response</h2>
+    <p>Control whether the firewall responds to ICMP echo requests (ping) on each interface. Default disabled for all interfaces.</p>
+
+    <div class="ping-control">
+      <h3>Internal Interface (192.168.1.1)</h3>
```
Courtney3141: These shouldn't be hardcoded - you must fetch them (unless this is an example...)

examples/firewall/ui_server.py
```diff
@@ -988,5 +1004,68 @@ def css(request):
...
+    </div>
+
+    <div class="ping-control">
+      <h3>External Interface (172.16.2.1)</h3>
+      <button id="external-enable">Enable Ping</button>
+      <button id="external-disable">Disable Ping</button>
+      <span id="external-status"></span>
+    </div>
+
+    <script>
```
Courtney3141, line 1048: I can see how this toggles enable/disable, but can't see how it retrieves what is the current setting?

include/lions/firewall/config.h
```diff
@@ -139,6 +141,7 @@ typedef struct fw_filter_config {
     region_resource_t internal_instances;
     region_resource_t external_instances;
     region_resource_t rule_id_bitmap;
+    fw_connection_resource_t icmp_module;
```
Courtney3141, line 153: As it is only the UDP filter that is using this resource currently (why not the ICMP filter btw?), I don't think it should go in the _general_ filter resource struct. However, if you think that most filters will need the connection excluding the TCP filter, then I'm happy to keep it here.

include/lions/firewall/icmp_helper.h
```diff
@@ -0,0 +1,120 @@
+/*
```
Courtney3141: I am thinking we can keep this in the `icmp.h` file.

include/lions/firewall/icmp.h
Courtney3141, line 116: Why have you created a separate `icmp_req_dest_t` struct for this type, but in other ICMP message types the data array is defined inside?

examples/firewall/icmp/icmp_module.c
```diff
@@ -27,116 +27,219 @@ net_client_config_t *net_configs[FW_NUM_INTERFACES] = {&ext_net_config, &int_net
...
+            ip_hdr->dst_ip = req->ip_hdr.src_ip;
+
+            /* Total length of ICMP echo reply IP packet */
+            uint16_t icmp_payload_len = req->echo.payload_len;
+            ip_hdr->tot_len = htons(IPV4_HDR_LEN_MIN + ICMP_ECHO_LEN);
+
+            /* Construct ICMP echo reply: 4 bytes (id + seq) + payload */
+            uint8_t *icmp_data = (uint8_t *)(pkt_vaddr + ICMP_HDR_OFFSET + ICMP_COMMON_HDR_LEN);
+            /* Echo identifier (network byte order) */
+            icmp_data[0] = (req->echo.echo_id >> 8) & 0xFF;
```
Courtney3141: I'm not so sure about this.

examples/firewall/icmp/icmp_module.c
```diff
@@ -27,116 +27,219 @@ net_client_config_t *net_configs[FW_NUM_INTERFACES] = {&ext_net_config, &int_net
...
+
+            /* Copy IP header */
+            memcpy(&icmp_time_exceeded->ip_hdr, &req->ip_hdr, IPV4_HDR_LEN_MIN);
+
+            /* Copy first bytes of data if applicable */
+            uint16_t to_copy_te = MIN(FW_ICMP_SRC_DATA_LEN, ntohs(req->ip_hdr.tot_len) - IPV4_HDR_LEN_MIN);
+            memcpy(&icmp_time_exceeded->data, req->time_exceeded.data, to_copy_te);
+            break;
+        case ICMP_REDIRECT_MSG:
+            sddf_printf("Redirect triggered");
```
Courtney3141: Remove this.

examples/firewall/docker/scripts/net_setup.sh
```diff
@@ -1,10 +1,21 @@
 # Copyright 2025, UNSW
 # SPDX-License-Identifier: BSD-2-Clause
 
-#!/bin/bash
+set -e
```
Courtney3141: What's going on here?

include/lions/firewall/filter.h
```diff
@@ -23,31 +23,45 @@ typedef enum {
...
-    FILTER_ACT_ALLOW = 1,
+	FILTER_ACT_ALLOW = 1,
     /* drop traffic */
-    FILTER_ACT_DROP = 2,
+	FILTER_ACT_DROP = 2,
     /* allow traffic, and additionally any return traffic */
     FILTER_ACT_CONNECT = 3,
     /* traffic is return traffic from a connect rule */
-    FILTER_ACT_ESTABLISHED
+    FILTER_ACT_ESTABLISHED = 4,
```
Courtney3141: Flip the order with established.

include/lions/firewall/filter.h
```diff
@@ -23,31 +23,45 @@ typedef enum {
...
 } fw_action_t;
 
-static const char *fw_filter_action_str[] = { "No rule", "Allow", "Drop", "Connect", "Established" };
+static const char *fw_filter_action_str[] = {
+    "No rule",
+    "Allow",
+    "Drop",
+    "Connect",
+    "Established",
+    "UDP Reject"
```
Courtney3141: Why "UDP Reject"

components/micropython/mpfirewallport.h
```diff
@@ -27,6 +27,7 @@ typedef struct fw_webserver_interface_state {
     fw_routing_table_t *routing_table;
     fw_filter_state_t filter_states[FW_MAX_FILTERS];
     uint16_t num_rules[FW_MAX_FILTERS];
+    bool ping_enabled;
```
Courtney3141, line 30: This is not shared memory.

examples/firewall/docker/scripts/net_setup.sh
```diff
@@ -25,30 +23,38 @@ ip link set tap1 master br1
```
Courtney3141, line 23: Should there be some additional tests written in `testing.sh` as well?

components/micropython/mpfirewallport.h
Courtney3141: Let's fix this later.

examples/firewall/filters/udp_filter.c
Courtney3141: Print still needs conditional if it is kept. Although I'm thinking we just remove it (it doesn't make sense anyway, there have been multiple packets received and processed, possibly with multiple actions).

examples/firewall/filters/udp_filter.c
Courtney3141: Missing indentation and full stop.

examples/firewall/filters/udp_filter.c
```diff
@@ -104,9 +116,25 @@ static void filter(void)
...
                 if (FW_DEBUG_OUTPUT) {
                     sddf_printf("%sUDP filter dropping via rule %u: (ip %s, port %u) -> (ip %s, port %u)\n",
-                                fw_frmt_str[filter_config.interface], rule_id,
-                                ipaddr_to_string(ip_hdr->src_ip, ip_addr_buf0), htons(udp_hdr->src_port),
-                                ipaddr_to_string(ip_hdr->dst_ip, ip_addr_buf1), htons(udp_hdr->dst_port));
+                        fw_frmt_str[filter_config.interface], rule_id,
+                        ipaddr_to_string(ip_hdr->src_ip, ip_addr_buf0), htons(udp_hdr->src_port),
+                        ipaddr_to_string(ip_hdr->dst_ip, ip_addr_buf1), htons(udp_hdr->dst_port));
+                }
+            } else if (action == FILTER_ACT_REJECT) {
```
Courtney3141: Maybe make this part of the previous if? i.e.

`if (action == FILTER_ACT_DROP || action == FILTER_ACT_REJECT)`

and only enqueue ICMP unreachable if reject? Just minimises repeated code.

examples/firewall/filters/icmp_filter.c
```diff
@@ -105,9 +117,27 @@ static void filter(void)
...
+                uint8_t icmp_type = icmp_hdr->type;
+                // An ICMP error message shouldn't be sent upon receival of anthor ICMP error message
+                if (icmp_is_error_message(icmp_type)) {
+                    err = net_enqueue_free(&rx_queue, buffer);
+                    assert(!err);
+                    returned = true;
+                } else {
+                    /* Enqueue an ICMP port unreachable message */
+                    err = enqueue_icmp_unreachable(buffer);
+                    assert(!err);
```
Courtney3141: I think this shouldn't be an assertion, as the queue could be full.

examples/firewall/filters/udp_filter.c
Courtney3141: Actually, I think we should make this a switch statement...

examples/firewall/icmp/icmp_module.c
```diff
@@ -27,116 +27,212 @@ net_client_config_t *net_configs[FW_NUM_INTERFACES] = {&ext_net_config, &int_net
 
 net_queue_handle_t net_queue[FW_NUM_INTERFACES];
 fw_queue_t icmp_queue[FW_NUM_INTERFACES];
+fw_queue_t filter_icmp_queue[FW_MAX_FILTERS];
+
+static bool process_icmp_request(icmp_req_t *req, uint8_t out_int, bool *transmitted)
```
Courtney3141, line 32: Isn't transmitted a global?

examples/firewall/icmp/icmp_module.c
```diff
@@ -27,116 +27,212 @@ net_client_config_t *net_configs[FW_NUM_INTERFACES] = {&ext_net_config, &int_net
...
+
+            /* Total length of ICMP destination unreachable IP packet */
+            uint16_t icmp_total_len = (uint16_t)(ICMP_COMMON_HDR_LEN + sizeof(icmp_echo_t) + req->echo.payload_len);
+            ip_hdr->tot_len = htons(IPV4_HDR_LEN_MIN + icmp_total_len);
+
+            /* Construct ICMP echo reply: 4 bytes (id + seq) */
+            icmp_echo_t *icmp_echo = (icmp_echo_t *)(pkt_vaddr + ICMP_ECHO_OFFSET);
+
+            /* Set Echo-specific fields from the request */
+            icmp_echo->id = htons(req->echo.echo_id);
```
Courtney3141, line 99: do we switch the byte order in the router too?

Courtney3141: Okay we do. Maybe we shouldn't bother switching the order twice if it is never modified?

include/lions/firewall/config.h
```diff
@@ -119,7 +119,9 @@ typedef struct fw_icmp_module_config {
     /* IP address of interfaces */
     uint32_t ips[FW_NUM_INTERFACES];
     fw_connection_resource_t routers[FW_NUM_INTERFACES];
+    fw_connection_resource_t filters[FW_MAX_FILTERS];
```
Courtney3141, line 126: filters need to be associated with an interface

examples/firewall/icmp/icmp_module.c
```diff
@@ -27,116 +27,212 @@ net_client_config_t *net_configs[FW_NUM_INTERFACES] = {&ext_net_config, &int_net
...
-            assert(!err);
 
-            /* Currently we only support destination unreachable ICMP traffic */
-            if (req.type != ICMP_DEST_UNREACHABLE) {
-                sddf_printf("ICMP module: Interface %u requested unsupported ICMP type %u!\n", out_int, req.type);
-                continue;
-            }
+    /* Process ICMP requests from filters */
+    for (uint8_t filter_idx = 0; filter_idx < icmp_config.num_filters; filter_idx++) {
+        uint8_t out_int = filter_idx;
```
Courtney3141: shouldn't this be output interface index?

examples/firewall/routing/routing.c
```diff
@@ -59,36 +59,140 @@ static bool tx_webserver; /* Packet has been transmitted to the webserver */
...
+    int err = icmp_enqueue_error(&icmp_queue, ICMP_DEST_UNREACHABLE, code, pkt_vaddr);
     if (!err) {
         notify_icmp = true;
     }
 
     return err;
 }
 
-static void transmit_packet(net_buff_desc_t buffer, uint8_t *mac_addr)
+static void enqueue_icmp_ping(net_buff_desc_t buffer)
```
Courtney3141: This function is repeated 3 times. Can we make `icmp_enqueue_echo_reply` return a `bool` and remove this repeated code?

examples/firewall/routing/routing.c
```diff
@@ -59,36 +59,140 @@ static bool tx_webserver; /* Packet has been transmitted to the webserver */
...
+        return out_int == ROUTING_OUT_EXTERNAL;
+    } else if (interface == FW_INTERNAL_INTERFACE_ID) {
+        return out_int == ROUTING_OUT_SELF;
+    } else {
+        return false;
+    }
+}
+
+static void check_enqueue_icmp_redirect(net_buff_desc_t buffer, fw_routing_interfaces_t out_int, uint32_t next_hop) {
+    sddf_printf("Enetred redirect check");
```
Courtney3141: Remove

examples/firewall/routing/routing.c
```diff
@@ -59,36 +59,140 @@ static bool tx_webserver; /* Packet has been transmitted to the webserver */
...
+    if (interface == FW_EXTERNAL_INTERFACE_ID) {
+        return out_int == ROUTING_OUT_EXTERNAL;
+    } else if (interface == FW_INTERNAL_INTERFACE_ID) {
+        return out_int == ROUTING_OUT_SELF;
+    } else {
+        return false;
+    }
+}
+
+static void check_enqueue_icmp_redirect(net_buff_desc_t buffer, fw_routing_interfaces_t out_int, uint32_t next_hop) {
```
Courtney3141: Maybe worth waiting until router combination before adding this functionality since it can't be tested yet.

examples/firewall/routing/routing.c
```diff
@@ -393,9 +579,13 @@ void notified(microkit_channel ch)
          * This is the channel between the ARP component and the
          * routing component
          */
+        sddf_printf("%sROUTING LOG: Notified by ARP module\n",
```
Courtney3141: Don't need this currently.

examples/firewall/routing/routing.c
```diff
@@ -393,9 +579,13 @@ void notified(microkit_channel ch)
...
+            fw_frmt_str[router_config.interface]);
         process_arp_waiting();
     } else {
         /* Router has been notified by a filter */
+        sddf_printf("%sROUTING LOG: Notified by filter module\n",
```
Courtney3141: Don't need this currently.

include/lions/firewall/icmp.h
```diff
@@ -52,44 +62,272 @@ typedef struct __attribute__((__packed__)) icmp_hdr {
...
+    /* IP Address of the new path */
+    uint32_t gateway_ip;
+    /* IP header of source packet */
+    ipv4_hdr_t ip_hdr;
+    /* First 8 bytes of data from source packet */
+    uint8_t data[FW_ICMP_SRC_DATA_LEN];
+} icmp_redirect_t;
+
+#define ICMP_REDIRECT_LEN (ICMP_COMMON_HDR_LEN + sizeof(icmp_redirect_t))
+#define ICMP_REDIRECT_OFFSET ICMP_PAYLOAD_OFFSET
```
Courtney3141: If they are all the same, probably no need to redefine for each ICMP type.

examples/firewall/icmp/icmp_module.c
```diff
@@ -27,116 +27,212 @@ net_client_config_t *net_configs[FW_NUM_INTERFACES] = {&ext_net_config, &int_net
...
+
+            /* Construct ICMP echo reply: 4 bytes (id + seq) */
+            icmp_echo_t *icmp_echo = (icmp_echo_t *)(pkt_vaddr + ICMP_ECHO_OFFSET);
+
+            /* Set Echo-specific fields from the request */
+            icmp_echo->id = htons(req->echo.echo_id);
+            icmp_echo->seq = htons(req->echo.echo_seq);
+
+            /* Copy the actual Echo payload data (the 'ping' data) */
+            uint8_t *echo_payload_dst = (uint8_t *)(icmp_echo + 1);
```
Courtney3141: +1?

include/lions/firewall/icmp.h
```diff
@@ -52,44 +62,272 @@ typedef struct __attribute__((__packed__)) icmp_hdr {
...
+
+/* ----------------- 8 - Echo Request / 0 - Echo Reply ---------------------------*/
+typedef struct __attribute__((__packed__)) icmp_echo {
+    /* Identifier to match requests with replies */
+    uint16_t id;
+    /* Sequence number */
+    uint16_t seq;
+    /* Payload data follows */
```
Courtney3141, line 116: Missing data field?

examples/firewall/icmp/icmp_module.c
```diff
@@ -27,116 +27,212 @@ net_client_config_t *net_configs[FW_NUM_INTERFACES] = {&ext_net_config, &int_net
...
+            /* Unused must be set to 0, as well as optional fields we are not currently using */
+            icmp_dest->unused = 0;
+            icmp_dest->len = 0;
+            icmp_dest->nexthop_mtu = 0;
+
+            /* Copy IP header */
+            memcpy(&icmp_dest->ip_hdr, &req->ip_hdr, IPV4_HDR_LEN_MIN);
+
+            /* Copy first bytes of data if applicable */
+            memcpy(&icmp_dest->data, req->dest.data, to_copy);
```
Courtney3141, line 126: Rather than +1 above ^, do something like this.

examples/firewall/ui_server.py
```diff
@@ -29,6 +29,7 @@
 OSErrOutOfMemory = 11
 OSErrInternalError = 12
 OSErrInvalidInput = 13
+OSErrInvalidActionTcp = 14
```
Courtney3141: Let's talk about this.

examples/firewall/filters/udp_filter.c
```diff
@@ -104,9 +116,25 @@ static void filter(void)
...
+                /* Enqueue an ICMP port unreachable message */
+                err = enqueue_icmp_unreachable(buffer);
+                assert(!err);
```
Courtney3141: These should not be asserts.

Issue #191, opened 2025-08-30
There should only be one Firewall system routing component


Courtney3141:
Currently there is one routing component per network interface which receives traffic from one NIC and routes out the other. This prevents traffic that is received from one NIC being routed out the same NIC. There should only be one routing component per system, and therefore one routing table.

PR #256, opened 2026-01-12
Mac address translation

ChronoComplexity:
Translations between broadcast and multicast IP addresses to their corresponding broadcast and multicast MAC addresses provided.

### DIFF
--- a/examples/firewall/routing/routing.c
+++ b/examples/firewall/routing/routing.c
@@ -60,11 +60,10 @@ static bool returned; /* Buffer has been returned to the rx virtualiser */
 static bool notify_arp; /* Arp request has been enqueued */
 static bool notify_icmp; /* Request has been enqueued to ICMP module */
 
-/* Masks for checking whether it is a broadcast address or not*/
+/* Masks for checking whether it is a broadcast address or not */
 #define MULTICAST_IP_MASK 0xf0000000
-#define MULTICAST_IP_NETWORK_ADDR 0xe0000000
+#define MULTICAST_IP_ADDR 0xe0000000
 #define BROADCAST_IP_ADDR 0xffffffff
-#define MULTICAST_MAC_SUFFIX_MASK 0x7fffff
 const uint8_t broadcast_mac_addr[] = {0xff, 0xff, 0xff, 0xff, 0xff, 0xff};
 
 /* Enqueue a request to the ICMP module to transmit a destination unreachable
@@ -101,10 +100,7 @@ static void transmit_packet(net_buff_desc_t buffer, uint8_t *mac_addr)
     eth_hdr_t *eth_hdr = (eth_hdr_t *)pkt_vaddr;
     ipv4_hdr_t *ip_hdr = (ipv4_hdr_t *)(pkt_vaddr + IPV4_HDR_OFFSET);
 
-    /* Replaces mac address if needed */
-    if (mac_addr != NULL) {
-        memcpy(&eth_hdr->ethdst_addr, mac_addr, ETH_HWADDR_LEN);
-    }
+    memcpy(&eth_hdr->ethdst_addr, mac_addr, ETH_HWADDR_LEN);
     memcpy(&eth_hdr->ethsrc_addr, router_config.mac_addr, ETH_HWADDR_LEN);
 
     /* Transmit packet out the NIC */
@@ -206,11 +202,23 @@ static void route(void)
                             buffer.io_or_offset / NET_BUFFER_SIZE);
             }
 
-            if (ip_hdr->dst_ip == BROADCAST_IP_ADDR) {
-                /* Limited broadcast, drop packet explicitly */
-                continue;
-            } else if ((ip_hdr->dst_ip & MULTICAST_IP_MASK) == MULTICAST_IP_NETWORK_ADDR) {
-                /* Multicast addresses not handled by router currently, skip */
+            /**
+             * Broadcast traffic should not be transmitted across subnets or
+             * retransmitted, thus it is explicitly dropped. Multicast traffic
+             * is not currently handled by the firewall.
+             */
+            if (ip_hdr->dst_ip == BROADCAST_IP_ADDR ||
+                !memcmp(eth_hdr->ethdst_addr, broadcast_mac_addr, ETH_HWADDR_LEN) ||
+                (ip_hdr->dst_ip & MULTICAST_IP_MASK) == MULTICAST_IP_ADDR) {
+
+                if (FW_DEBUG_OUTPUT) {
+                    sddf_printf("%sRouter received a broadcast MAC or multicast IP packet, dropping!\n",
+                                fw_frmt_str[router_config.interface]);
+                }
+
+                err = fw_enqueue(&rx_free, &buffer);
+                assert(!err);
+                returned = true;
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


Courtney3141:
After looking into how multicast IP traffic works, at this stage I don't think it is worth us implementing it. This is because:

1. Proper multicast management in a router would require the router to *subscribe* to all the multicast groups it wishes to propagate. To do this, the router would need support generating and listening for IGMP traffic.
2. Forwarding multicast traffic is not as simple as just transmitting any multicast traffic received on one interface out the other. The router needs to keep track of which hosts are subscribed to which multicast groups. When multicast traffic is received, it is only transmitted out interfaces that are known to have hosts subscribed to the group.
3. Multicast routing is also implemented differently to unicast, in that in unicast routing, traffic is routed to the destination IP, while multicast traffic is routed _away_ from the source IP, see [here](https://en.wikipedia.org/wiki/Reverse-path_forwarding).

For these reasons, I propose that we hold off on supporting multicast routing until we have completed [merging the two routing components](https://github.com/au-ts/lionsos/issues/191), as much of the above is meaningless until the router is capable of routing out of more than one interface.

Additionally, implementing multicast is complex enough (requiring additional data structures, as well as IGMP support and snooping) that it deserves a separate PR. I will make an issue for multicast traffic and IGMP support to return to later.

So for the time being, please remove the multicast parts of this PR. I will also leave some additional comments about the issues we discussed on Friday.

examples/firewall/routing/routing.c
```diff
@@ -264,58 +270,70 @@ static void route(void)
...
-                (pkt_waiting_full(&pkt_waiting_queue) &&
-                (arp == NULL || arp->state == ARP_STATE_PENDING)) ||
-                (arp == NULL && fw_queue_full(&arp_req_queue))) {
-
-                if (arp != NULL && arp->state == ARP_STATE_UNREACHABLE) {
-                    int icmp_err = enqueue_icmp_unreachable(buffer);
-                    if (icmp_err) {
-                        sddf_dprintf("%sROUTING LOG: Could not enqueue ICMP unreachable!\n",
+            /* Finds the MAC address, then transmits a packet */
+            if (next_hop == BROADCAST_IP_ADDR) {
```
Courtney3141: The check for broadcast traffic should occur prior to the destination IP being searched in the routing table, i.e. the call of `fw_routing_find_route`

examples/firewall/routing/routing.c
```diff
@@ -264,58 +270,70 @@ static void route(void)
...
+                /* valid arp entry found, transmit packet */
+                uint8_t broadcast_mac_addr[] = {0xff, 0xff, 0xff, 0xff, 0xff, 0xff};
```
Courtney3141: As I mentioned earlier, there is no reason why this fixed MAC address could not be defined as a `const` with the other hardcoded addresses at the top of the file (since it is never written to throughout the program).

I am also unsure of your comment here, as in fact there should _never_ be an ARP entry for the broadcast IP/MAC addresses, nor are you using an ARP entry

However, after looking into it further it seems that __all__ IP packets addressed to the broadcast IP address will __always__ have the broadcast MAC address as their destination MAC address in the encapsulating Ethernet header. So it seems that this MAC address substitution will be overwriting what is already written with the same value.

I note that in the current implementation of `transmit_packet` though, I never considered the case where the MAC does not need to be overwritten. So you have two choices:

A) Keep the code as is, but move the hardcoded broadcast MAC declaration to the same place as the broadcast IP macro.

B) Change the implementation of `transmit_packet` so that overwriting the MAC is optional. This could potentially be done by permitting the MAC address pointer to be `NULL`, and in this case the destination MAC is not modified.

I would prefer option B, but I will leave the choice to you.

Courtney3141:
Forgot to mention one other thing - please make sure you test your changes (i.e. is broadcast traffic propagated through the firewall now)

examples/firewall/routing/routing.c
```diff
@@ -261,7 +278,6 @@ static void route(void)
                 }
 
                 continue;
-
```
Courtney3141: Please remove whitespace diff.

examples/firewall/routing/routing.c
```diff
@@ -292,7 +308,7 @@ static void route(void)
             and send ARP request or await ARP response */
             if (arp == NULL || arp->state == ARP_STATE_PENDING) {
                 pkt_waiting_node_t *root = pkt_waiting_find_node(&pkt_waiting_queue,
-                                                                 next_hop);
```
Courtney3141: Please remove whitespace diff.

examples/firewall/routing/routing.c
```diff
@@ -201,6 +210,14 @@ static void route(void)
                             buffer.io_or_offset/NET_BUFFER_SIZE);
             }
 
+            if (ip_hdr->dst_ip == BROADCAST_IP_ADDR) {
```
Courtney3141: This is good, but you are missing something - as well as checking for the broadcast IP address 255.255.255.255, you also need to check whether the destination IP address is the broadcast IP address for any of the subnets you are aware of.

For example, if you are transmitting out 192.168.2.0/24, the broadcast IP is 192.168.2.255. Typically, packets destined for the _subnet_ broadcast address from other subnets are not forwarded, so it's the responsibility of the router to check for this.

So what I propose is creating a function similar to `fw_routing_find_route`, like `fw_routing_check_broadcast` which checks if the destination IP address is either 255.255.255.255, or a subnet broadcast IP for one of the subnets you are aware of (by reading the routing table), and if so, drops the packet (the resultant behaviour could later be configured).

In the future, this function could also be extended for handling multicast as well. For the time being you could move this multicast `else if` into function as is.

ChronoComplexity: The subnet broadcast should be under the subnet as calculated in fw_routing_find_route with how it's currently done I think.

ChronoComplexity: I'm considering combining it with find_route tbh.

ChronoComplexity: Wait quick q, how is this working:
  } else if (match->interface == ROUTING_OUT_EXTERNAL &&
               match->next_hop == FW_ROUTING_NONEXTHOP) {
        *next_hop = ip;
        *interface = ROUTING_OUT_EXTERNAL;
shouldn't it be like *next_hop = entry->ip; or *next_hop = entry->next_hop.

examples/firewall/routing/routing.c
```diff
@@ -60,6 +60,13 @@ static bool returned; /* Buffer has been returned to the rx virtualiser */
 static bool notify_arp; /* Arp request has been enqueued */
 static bool notify_icmp; /* Request has been enqueued to ICMP module */
 
+/* Masks for checking whether it is a broadcast address or not*/
```
Courtney3141: There needs to be a space here: "not */"

examples/firewall/routing/routing.c
```diff
@@ -94,7 +101,10 @@ static void transmit_packet(net_buff_desc_t buffer, uint8_t *mac_addr)
     eth_hdr_t *eth_hdr = (eth_hdr_t *)pkt_vaddr;
     ipv4_hdr_t *ip_hdr = (ipv4_hdr_t *)(pkt_vaddr + IPV4_HDR_OFFSET);
 
-    memcpy(&eth_hdr->ethdst_addr, mac_addr, ETH_HWADDR_LEN);
+    /* Replaces mac address if needed */
```
Courtney3141: We no longer need this functionality right? So this can be removed.

PR #258, opened 2026-01-16
trivial(firewall): improve metaprogram structure error handling

debater-coder:
Previously, if calculating the size of a datastructure failed, it would throw an opaque error message which does not inform what error occured.

### DIFF
--- a/examples/firewall/meta.py
+++ b/examples/firewall/meta.py
@@ -1040,8 +1040,7 @@ if __name__ == "__main__":
                 # Data structure size has already been calculated
                 continue
             try:
-                import os
-                if not os.path.exists(structure.elf_name):
+                if not path.exists(structure.elf_name):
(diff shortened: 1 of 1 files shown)


examples/firewall/meta.py
```diff
@@ -1040,6 +1040,9 @@ def generate(sdf_file: str, output_dir: str, dtb: DeviceTree):
                 # Data structure size has already been calculated
                 continue
             try:
+                import os
+                if not os.path.exists(structure.elf_name):
```
Ivan-Velickovic:
```suggestion
                if not path.exists(structure.elf_name):
```

Forgot to mention, I did this just while I was testing things out of the script, looks like `from os import path` already exists at the top of the file.

Ivan-Velickovic:
Thanks!

PR #259, opened 2026-01-16
Firewall: Implement NAT 

### LINKED ISSUE #188: Support NAT in the Firewall
The firewall system needs to support NAT and port forwarding.

debater-coder:
Closes #188 .

This PR adds Network Address Translation (NAT) support to the firewall example. Implementing NAT allows the firewall to be used in scenarios where it is used as a gateway to the public internet.

### Separate PDs per protocol per filter
This implementation uses separate PDs to perform address translation. For each NIC, there are as many NAT components as there are supported protocols (currently only UDP). This means that the PD sits in between each of the filters (since there is one filter per protocol) and the router.

This also better aligns with the LionsOS principles:
> As far as is practical, operating systems components do a single thing.

### Transparent to filter and router PDs
The NAT is transparent to both the router and the filters (by using `fw_queue` connected to the filter and router). This means that NAT can be selectively enabled for different protocols, or it can be completely disabled (with a direct connection between filters and router). This will mean that usecases that do not require NAT incur zero overhead.

### NAT connection table using shared memory
For every NAT PD, there is a port table used to map ephemeral ports to the original port and IP. This allows for returning traffic to be send to its original sender. The table is indexed by ephemeral port (including a base offset for index 0). Thus, every protocol has its own pool of ephemeral ports. This table may eventually need to store a `last_seen` timestamp and the state of the TCP connection (for the TCP network address translator).

### Endpoint-Independent Mapping
Currently, this PR implements Endpoint-Independent Mapping (EIM) NAT. This means that upon translation, the source port assigned to the packet is only dependent on the original _source_ IP address and port of the packet (rather than also being dependent on the _destination_). Some NAT implementations create a different mapping based on both the source and destination hosts as a means of basic filtering (Symmetric NAT). However, the filtering modules already provide stateful firewall functionality, so I have chosen to avoid duplicating this and having the NAT modules focus solely on address and port translation.

However, this approach limits the number of translatable concurrent connections to the number of ports (65535). On larger networks Symmetric NAT becomes more useful, as multiple connections can be assigned to the same external port, distinguishing between them based on the destination IP and port. However, this approach has the disadvantage of needing to do additional searching within the NAT mappings (this current implementation allows for the original source IP and port corresponding to an ephemeral port to be found in a single array indexing operation). 

Since the Symmetric NAT approach would be slower (due to additional searching) and more memory intensive (due to larger search structures) it is best suited to be implemented in a separate component, to avoid unnecessary overhead for smaller networks. These implementations could be switched between using the metaprogram.   

## Testing
Both UDP and TCP outgoing and returning traffic have been confirmed to work using the test program in `test_nat.py`. This test program prints the IP and port of packets received by the test server, and sends the string `return traffic` back to the client, demonstrating both the initial translation and the reversal upon returning traffic.

Changing NAT settings in the web interface has also been tested. 

image

image

## Tasks
- [x] NAT for UDP outgoing traffic
- [x] Handle returning traffic
- [x] NAT for TCP traffic
- [x] Freelist for NAT port-mapping (allows for port-mapping entries to be reused)
- [x] Remove NAT mappings on timeout
- [x] Generic NAT implementation
- [x] Configuration in the web
    - Source NAT IP
    - TCP Timeout
    - UDP Timeout

## Future improvements
- #261
- #262
- #265
- #264
- #263

### DIFF
--- a/examples/firewall/ui_server.py
+++ b/examples/firewall/ui_server.py
@@ -459,7 +459,7 @@ def set_snat_ip(request, protocolStr, interfaceStr, ipStr):
             raise OSError(OSErrInvalidInput, OSErrStrings[OSErrInvalidInput])
         protocol = protocolNums[protocolStr]
 
-        lions_firewall.set_snat_ip(interface, protocol, ipToInt(ipStr) if ipStr != "" else 0)
+        lions_firewall.set_snat_ip(interface, protocol, ipToInt(ipStr) if ipStr != "null" else 0)
         return {"status": "ok"}
     except OSError as OSErr:
         print(f"UI SERVER|ERR: OS Error: set_snat_ip: {OSErrStrings[OSErr.errno]}")
@@ -477,7 +477,7 @@ def get_nat_timeout(request, protocolStr):
         protocol = protocolNums[protocolStr]
 
         timeout = lions_firewall.get_nat_timeout(protocol)
-        return {"snat": timeout / 10**9}
+        return {"timeout": timeout / 10**9}
     except OSError as OSErr:
         print(f"UI SERVER|ERR: OS Error: get_nat_timeout: {OSErrStrings[OSErr.errno]}")
         return {"error": OSErrStrings[OSErr.errno]}, 404
@@ -1054,7 +1054,7 @@ def rules(request):
 
 @app.route("/nat")
 def nat(request):
-    html = """
+    html = r"""
     <!DOCTYPE html>
     <html>
         <head>
@@ -1067,43 +1067,107 @@ def nat(request):
             <nav>
                 <a href="/">Home</a> | <a href="/routing_config">Routing Config</a> | <a href="/rules">Rules</a> | <a href="/interface">Interface</a> | <a href="/nat">Network Address Translation</a>
             </nav>
+            <div>
+                <label for="tcp-timeout">TCP mappings timeout</label>
+                <input
+                    type="number"
+                    name="tcp-timeout"
+                    id="tcp-timeout"
+                />
+                <button onClick="updateTimeout('tcp')">Update TCP Timeout</button>
+            </div>
+            <div>
+                <label for="udp-timeout">UDP mappings timeout</label>
+                <input
+                    type="number"
+                    name="udp-timeout"
+                    id="udp-timeout"
+                />
+                <button onClick="updateTimeout('udp')">Update UDP Timeout</button>
+            </div>
             <h2>Internal interface NAT</h2>
+            <h3>TCP</h3>
             <div>
                 <input
                     type="checkbox"
                     name="nat-enabled"
-                    id="internal-nat-enabled"
+                    id="internal-tcp-enabled"
+                    onchange="disabledHandler('tcp', 'internal')"
                 />
-                <label for="nat-enabled">NAT enabled?</label>
-                <button>Update NAT enabled</button>
-            </div>
-            <div>
+                <label for="nat-enabled">NAT enabled?</label><br>
                 <label for="nat-ip">Source NAT IP for this interface</label>
                 <input
                     type="text"
                     name="nat-ip"
-                    id="internal-nat-ip"
+                    id="internal-tcp-snat"
                 />
-                <button>Update NAT IP</button>
+                <button onClick="updateSnat('tcp', 'internal')">Update NAT settings</button>
             </div>
+            <h3>UDP</h3>
             <div>
-                <label for="tcp-timeout">Timeout for TCP mappings for this interface</label>
                 <input
-                    type="number"
-                    name="tcp-timeout"
-                    id="internal-tcp-timeout"
+                    type="checkbox"
+                    name="nat-enabled"
+                    id="internal-udp-enabled"
+                    onchange="disabledHandler('udp', 'internal')"
                 />
-                <button>Update TCP Timeout</button>
-            </div>
-            <div>
-                <label for="udp-timeout">Timeout for UDP mappings for this interface</label>
+                <label for="nat-enabled">NAT enabled?</label><br>
+                <label for="nat-ip">Source NAT IP for this interface</label>
                 <input
-                    type="number"
-                    name="udp-timeout"
-                    id="internal-udp-timeout"
+                    type="text"
+                    name="nat-ip"
+                    id="internal-udp-snat"
                 />
-                <button>Update UDP Timeout</button>
+                <button onClick="updateSnat('udp', 'internal')">Update NAT settings</button>
             </div>
+            <script>
+            const placeholderInterfaceMapping = {
+                "internal": 0,
+                "external": 1
+            };
+
+            const getTimeout = async (protocol) => {
+                const data = await fetch(`/api/nat/${protocol}/timeout`).then(res => res.json());
+                if (data?.timeout) {
+                    document.querySelector(`#${protocol}-timeout`).value = data.timeout;
+                }
+            };
+
+            const getSnat = async (protocol, interface) => {
+                const data = await fetch(`/api/nat/${protocol}/${interface}/snat`).then(res => res.json());
+                document.querySelector(`#${interface}-${protocol}-enabled`).checked = !!data?.snat;
+                if (data?.snat) {
+                    document.querySelector(`#${interface}-${protocol}-snat`).value = data.snat;
+                }
+                document.querySelector(`#${interface}-${protocol}-snat`).placeholder = (await fetch(`/api/interfaces/${placeholderInterfaceMapping[interface]}`).then(res => res.json())).ip;
+            }
+
+            const updateTimeout = async (protocol) => {
+                const timeout = document.querySelector(`#${protocol}-timeout`).value;
+                await fetch(`/api/nat/${protocol}/timeout/${timeout}`, {method: "PUT"});
+                await getTimeout(protocol);
+            };
+
+            const updateSnat = async (protocol, interface) => {
+                const enabled = document.querySelector(`#${interface}-${protocol}-enabled`).checked;
+                const snat = enabled ? document.querySelector(`#${interface}-${protocol}-snat`).value : "null";
+                await fetch(`/api/nat/${protocol}/${interface}/snat/${snat}`, {method: "PUT"});
+                await getSnat(protocol, interface);
+            }
+
+            const disabledHandler = (protocol, interface) => {
+                document.querySelector(`#${interface}-${protocol}-snat`).disabled = ! document.querySelector(`#${interface}-${protocol}-enabled`).checked;
+            }
+
+            window.onload = () => {
+                getTimeout("udp");
+                getTimeout("tcp");
+                getSnat("udp", "internal");
+                getSnat("tcp", "internal");
+                disabledHandler("udp", "internal");
+                disabledHandler("tcp", "internal");
+            }
+            </script>
         </body>
     </html>
     """


debater-coder:
Hi @Courtney3141. I would appreciate it if you could take a look at this. I'd be happy to clean up if there are any blocking issues with this.
Thanks

Courtney3141:
Hey! I definitely intend to, sorry for the delay but I have been ill last week and backed up on other's PRs that were opened before this. Hoping I'll get it done this week, if not definitely next!

examples/firewall/nat/nat.c
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+void notified(microkit_channel ch)
+{
+    if (ch == nat_config.filter.ch) {
+        translate();
+    } else if (ch == timer_config.driver_id) {
+        uint64_t now = sddf_timer_time_now(timer_config.driver_id);
+        fw_nat_free_expired_mappings(nat_interface_config, port_table, webserver_state->timeout, now);
```
Courtney3141, line 145: Generally better to pass a pointer to a struct (`nat_interface_config`) to avoid overhead from copying.

examples/firewall/nat/nat.c
```diff
@@ -0,0 +1,180 @@
...
+
+/* Incoming packets from filter */
+fw_queue_t filter_queue;
+/* Outgoing packets to router */
+fw_queue_t router_queue;
+/* Virtual address of rx buffer data region */
+uintptr_t data_vaddr;
+/* Table storing ephemeral ports */
+fw_nat_port_table_t *port_table;
+fw_nat_interface_config_t nat_interface_config;
```
Courtney3141, line 33: This can be stored/used as a pointer.

examples/firewall/nat/nat.c
```diff
@@ -0,0 +1,180 @@
...
+
+        uintptr_t pkt_vaddr = data_vaddr + buffer.io_or_offset;
+        ipv4_hdr_t *ip_hdr = (ipv4_hdr_t *)(pkt_vaddr + IPV4_HDR_OFFSET);
+        char *transport_hdr = (char *)(pkt_vaddr + transport_layer_offset(ip_hdr));
+
+        uint16_t *src_port = (uint16_t*)(transport_hdr + nat_config.src_port_off);
+        uint16_t *dst_port = (uint16_t*)(transport_hdr + nat_config.dst_port_off);
+        uint16_t *check = (uint16_t*)(transport_hdr + nat_config.check_off);
+
+        uint32_t snat = webserver_state->interfaces[nat_config.interface].snat;
```
Courtney3141, line 70: This shouldn't be changed while `translate` is executing, (since a notification from the webserver will not be received), thus could be moved outside the `while` loop (line 56) for simplicity. Although not critical.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+ */
+typedef struct fw_nat_port_mapping fw_nat_port_mapping_t;
+struct fw_nat_port_mapping {
+    /* original source ip of traffic */
+    uint32_t src_ip;
+    /* original source port of traffic (network byte order) */
+    uint16_t src_port;
+    /* next free node (for nodes in free list only) */
+    fw_nat_port_mapping_t *next_free;
+    bool is_valid;
```
Courtney3141, line 46: This can likely be eliminated by using an invalid IP for port mapping, e.g. 255.255.255.255 or 0

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+} fw_nat_port_table_t;
+
+/**
+ * Find the destination IP address and port for an incoming packet.
+ * If the destination IP address matches the source NAT IP address
+ * of the NAT on another interface, then the packet corresponds
+ * to returning traffic and the port mapping corresponding
+ * to that ephemeral port will be returned.
+ *
+ * @param interfaces configuration of NAT for each interface
```
Courtney3141, line 68: Please update parameter descriptions to match new parameters.

examples/firewall/nat/nat.c
```diff
@@ -0,0 +1,180 @@
...
+                    sddf_printf("%s%s NAT LOG: translated to %s:%u\n", fw_frmt_str[nat_config.interface],
+                                ipv4_proto_name(nat_config.protocol),
+                                ipaddr_to_string(snat, ip_addr_buf0),
+                                htons(*src_port));
+                }
+            } else {
+                sddf_printf("%s%s NAT LOG: ephemeral ports ran out!\n", fw_frmt_str[nat_config.interface],
+                            ipv4_proto_name(nat_config.protocol));
+            }
+        } else {
```
Courtney3141, line 112: I'm thinking this should only print when `snat == 0`, and not when `dst_ip == nat_interface_config.ip`

examples/firewall/nat/nat.c
```diff
@@ -0,0 +1,180 @@
...
+                           );
+            }
+            *dst_port = dst_mapping->src_port;
+            ip_hdr->dst_ip = dst_mapping->src_ip;
+            ip_hdr->check = 0;
+
+            recalculate_checksum = true;
+        }
+
+        if (snat && ip_hdr->dst_ip != nat_interface_config.ip) {
```
Courtney3141, line 92: I think we may eventually have to handle the case where we receive traffic for the firewall's IP on another interface (rather than just _this_ interface), but this can be handled later...

examples/firewall/nat/nat.c
```diff
@@ -0,0 +1,180 @@
...
+                recalculate_checksum = true;
+
+                if (FW_DEBUG_OUTPUT) {
+                    sddf_printf("%s%s NAT LOG: translated to %s:%u\n", fw_frmt_str[nat_config.interface],
+                                ipv4_proto_name(nat_config.protocol),
+                                ipaddr_to_string(snat, ip_addr_buf0),
+                                htons(*src_port));
+                }
+            } else {
+                sddf_printf("%s%s NAT LOG: ephemeral ports ran out!\n", fw_frmt_str[nat_config.interface],
```
Courtney3141, line 109: I *think* the packet should be freed/dropped here, rather than transmitting through anyway... Or at least this could be configurable.

Courtney3141: Note this will require Rx free queues with the Rx virtualiser, as the filters/routers have.

components/micropython/mpfirewallport.h
```diff
@@ -29,7 +30,12 @@ typedef struct fw_webserver_interface_state {
     uint16_t num_rules[FW_MAX_FILTERS];
 } fw_webserver_interface_state_t;
 
-extern fw_webserver_interface_state_t webserver_state[FW_NUM_INTERFACES];
+typedef struct fw_webserver_state {
+    fw_webserver_interface_state_t interfaces[FW_NUM_INTERFACES];
+    fw_nat_webserver_state_t *nat_state[FW_MAX_NAT];
```
Courtney3141: Since `fw_nat_webserver_state_t` already contains an array of interface state (`fw_nat_webserver_interface_state_t`), I'm thinking this should just be a single pointer, rather than an array of pointers.

include/lions/firewall/config.h
```diff
@@ -16,6 +17,7 @@
 
 #define FW_MAX_FW_CLIENTS 61
 #define FW_MAX_FILTERS 61
+#define FW_MAX_NAT 16
```
Courtney3141, line 20: `FW_MAX_FILTERS` is a per interface constant. I'm wondering if you intended `FW_MAX_NAT` to be the same. In particular, I would imagine from your description that the maximum number of NAT components per interface should be equal to the maximum number of filters (in which case no need to define a new macro).

include/lions/firewall/config.h
```diff
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include "sddf/serial/queue.h"
```
Courtney3141, line 9: Don't think this is needed.

components/micropython/mpfirewallport.c
```diff
@@ -210,10 +211,17 @@ void mpfirewall_handle_notify(void) {
...
-            webserver_state[i].filter_states[j].rule_table = fw_config.interfaces[i].filters[j].rules.vaddr;
+            webserver_state.interfaces[i].filter_states[j].rule_table = fw_config.interfaces[i].filters[j].rules.vaddr;
         }
     }
+
+    for (uint8_t i = 0; i < fw_config.num_nat_state; i++) {
+        webserver_state.nat_state[i] = fw_config.nat_state[i].region.vaddr;
+
+        /* Set sensible NAT timeout defaults */
+        webserver_state.nat_state[i]->timeout = 30 * NS_IN_S;
```
Courtney3141, line 225: Default value should be configured in metaprogram.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * Find the ephemeral port to use for a source IP and port.
+ * Attempts to reuse an existing mapping for that IP and port,
+ * only creating a new entry if not found.
+ *
+ * @param config NAT config for this interface
+ * @param ports ephemeral port table for this interface
+ * @param src_ip source IP in network byte order
+ * @param src_port source port in network byte order
+ *
```
Courtney3141, line 104: add `now` parameter

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+    fw_nat_port_mapping_t *next_free;
+    bool is_valid;
+    /* sddf timer timestamp (nanoseconds) for last time a packet was sent or received on this port */
+    uint64_t last_used_ts;
+};
+
+typedef struct fw_nat_port_table {
+    /* number of valid NAT entries */
+    uint16_t size;
+    /* largest initialised entry in the NAT table (could be valid or free) */
```
Courtney3141, line 54: Do you mean _allocated_ or free?

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+ * Find the ephemeral port to use for a source IP and port.
+ * Attempts to reuse an existing mapping for that IP and port,
+ * only creating a new entry if not found.
+ *
+ * @param config NAT config for this interface
+ * @param ports ephemeral port table for this interface
+ * @param src_ip source IP in network byte order
+ * @param src_port source port in network byte order
+ *
+ * @returns ephemeral port in network byte order.
```
Courtney3141, line 105: Or 0 upon failure.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+        ports->free_head = mapping->next_free;
+
+        mapping->src_port = src_port;
+        mapping->src_ip = src_ip;
+        mapping->last_used_ts = now;
+        ports->mappings[ports->largest_index].is_valid = true;
+        return htons(config.base_port + mapping - ports->mappings);
+    }
+
+    if (ports->size >= config.ports_capacity) {
```
Courtney3141, line 133: If the `free_head == NULL` here, `ports->size >= config.ports_capacity` should _always_ be true, right? Otherwise something has gone wrong. So, maybe rather than making this a separate `if`, you can add an `else` to `if (ports->free_head)`, and _assert_ that `size >= capacity`.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+
+        /* Remove from front of list */
+        fw_nat_port_mapping_t *mapping = ports->free_head;
+        ports->free_head = mapping->next_free;
+
+        mapping->src_port = src_port;
+        mapping->src_ip = src_ip;
+        mapping->last_used_ts = now;
+        ports->mappings[ports->largest_index].is_valid = true;
+        return htons(config.base_port + mapping - ports->mappings);
```
Courtney3141, line 130: Although I believe this pointer arithmetic _is_ correct, it is slightly confusing to read, and prone to errors. Storing `free_head` as an index rather than pointer will make this cleaner.

include/lions/firewall/nat.h
Courtney3141: Okay, I think I see what you are doing here. `largest_index` means the largest port number index that _has_ been allocated. This would be an optimisation for the case where the port mapping table never gets full. I am happy to keep this in, but be careful to use it in the right way.

If you add some additional book-keeping, this could instead be the highest _allocated_ mapping slot (namely, you could reuse the `next_free` index so that when the mapping has been allocated, it points to the previous highest allocated slot, e.g.

Image

So, when 5 is freed, we know immediately that 3 is the next highest index. Thus highest index can be updated. And your free list can be maintained fairly easily in numerically order. So when a port number is reallocated, we just allocated from the lowest index (only updating the highest index if it is larger.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+ * This is an endpoint independent mapping since only source address and port are used.
+ */
+typedef struct fw_nat_port_mapping fw_nat_port_mapping_t;
+struct fw_nat_port_mapping {
+    /* original source ip of traffic */
+    uint32_t src_ip;
+    /* original source port of traffic (network byte order) */
+    uint16_t src_port;
+    /* next free node (for nodes in free list only) */
+    fw_nat_port_mapping_t *next_free;
```
Courtney3141, line 45: Since with our port mappings, what we really care about are the indices into the table (i.e. the port that will be allocated), I believe it is more optimal to store _indices_ to port mappings, rather than pointers. This has additional benefits like a notable reduction of memory (`uint16_t` vs `uint64_t`) as well as other things that I will point out.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+    uint64_t last_used_ts;
+};
+
+typedef struct fw_nat_port_table {
+    /* number of valid NAT entries */
+    uint16_t size;
+    /* largest initialised entry in the NAT table (could be valid or free) */
+    uint16_t largest_index;
+    /* head of free nodes */
+    fw_nat_port_mapping_t *free_head;
```
Courtney3141, line 57: As discussed above, we can change this to a `uint16_t` index of the first free mapping slot.

include/lions/firewall/nat.h
Courtney3141: Additionally, rather than building up these next pointers on the fly, I think at `init` time you should initialise the port mapping list so that the `next_free` points to the following index in the table. Then the capacity of the pot mapping table can be used to represent an invalid pointer rather than `NULL`

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+ * @param src_ip source IP in network byte order
+ * @param src_port source port in network byte order
+ *
+ * @returns ephemeral port in network byte order.
+ */
+static inline uint16_t fw_nat_find_ephemeral_port(fw_nat_interface_config_t config, fw_nat_port_table_t *ports,
+                                                  uint32_t src_ip, uint16_t src_port, uint64_t now)
+{
+    /* Search for an existing mapping */
+    for (uint16_t i = 0; i < ports->size; i++) {
```
Courtney3141, line 111: As discussed above, we can use `ports->highest_index` rather than `ports->size` here as an optimisation.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+        ports->size++;
+
+        /* Remove from front of list */
+        fw_nat_port_mapping_t *mapping = ports->free_head;
+        ports->free_head = mapping->next_free;
+
+        mapping->src_port = src_port;
+        mapping->src_ip = src_ip;
+        mapping->last_used_ts = now;
+        ports->mappings[ports->largest_index].is_valid = true;
```
Courtney3141, line 129: This _may_ be true, but is confusing to read, and I'm not certain it _is_ true. As mentioned, I think the free lists should be initialised at `init` time (with essentially 0 expensive).

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+        return htons(config.base_port + mapping - ports->mappings);
+    }
+
+    if (ports->size >= config.ports_capacity) {
+        return 0; /* Ephemeral ports pool is full */
+    }
+
+    /* Initialise a new entry in the NAT table and assign its ephemeral port */
```
Courtney3141, line 137: The control flow of this function is quite confusing. If the free list is pre-initialised this code would never be reached (thus the end of the function can just be the assertion I mentioned above, and a return of 0).

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+ * @param config NAT config for this interface
+ * @param ports ephemeral port table for this interface
+ * @param port ephemeral port to be freed in host byte order
+ */
+static inline void fw_nat_free_ephemeral_port(fw_nat_interface_config_t config, fw_nat_port_table_t *ports,
+                                              uint16_t port)
+{
+    fw_nat_port_mapping_t *mapping = &ports->mappings[port - config.base_port];
+
+    sddf_printf("NAT LOG: free head is now: %p\n", ports->free_head);
```
Courtney3141, line 159: Remove.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+    mapping->src_ip = 0;
```
Courtney3141, line 160: Setting everything to 0 is unnecessary. When it is re-allocated, these values will be overwritten. Only set the minimal field to 0 that is required to identify that the entry is invalid.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+    mapping->src_port = 0;
+    mapping->next_free = ports->free_head;
+    mapping->is_valid = false;
+    mapping->last_used_ts = 0;
+
+    /* this index is now the new head */
+    ports->free_head = mapping;
+    sddf_printf("NAT LOG: free head is now: %p\n", ports->free_head);
```
Courtney3141, line 168: remove.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+
+    ports->size--;
+}
+
+/**
+ * Frees all port mappings older than the timeout duration.
+ * @param config NAT config for this interface
```
Courtney3141, line 175: Newline here.

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+#include <os/sddf.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <sddf/util/util.h>
+#include <sddf/network/util.h>
+#include <lions/firewall/common.h>
+#include <lions/firewall/array_functions.h>
+#include <lions/firewall/config.h>
+
+#define NAT_TIMEOUT_INTERVAL_NS (5 * NS_IN_S)
```
Courtney3141, line 17: This should really be called something like timeout _tick_, since it's not necessarily the timeout...

include/lions/firewall/nat.h
```diff
@@ -0,0 +1,194 @@
...
+ * @param timeout duration in nanoseconds for which entries older than it will be freed
+ * @param now the time now as an SDDF timestamp
+ */
+static inline void fw_nat_free_expired_mappings(fw_nat_interface_config_t config, fw_nat_port_table_t *ports,
+                                                uint64_t timeout, uint64_t now)
+{
+    for (uint16_t i = 0; i < ports->largest_index; i++) {
+        fw_nat_port_mapping_t *mapping = &ports->mappings[i];
+
+        if (mapping->is_valid && now > timeout && mapping->last_used_ts <= now - timeout) {
```
Courtney3141, line 186: Isn't the condition to free if `now >=  mapping->last_used_ts + timeout`? why do you need two timing conditions here? Writing as above avoids thinking of any overflow cases.

include/lions/firewall/config.h
Courtney3141: I will address this in my [new branch](https://github.com/au-ts/lionsos/tree/fw_nat_by_interface)

components/micropython/mpfirewallport.c
Courtney3141: I will address this in my [new branch](https://github.com/au-ts/lionsos/tree/fw_nat_by_interface)

Courtney3141:
Heads up, *should* have fixed the build issues now (updated the sDDF dependency in another PR). But definitely will take longer than today to finish up my config change branch (busy day today!)

PR #270, opened 2026-02-11
Refactor the meta program.

cazb2:
This commit primarily refactors the meta program. The goal was to represent the various connections and components in python data classes to improve the readability of the meta program.

By representing these connections and components in data classes this commit supports generating a dot file which can be used to view schematics/topology of the design defined in the meta program.

In addition this commit changes the internal and external labelling to align with the underlying hardware. The internal and external naming is now just a label with the meta program now being better able to support more than two interfaces.

In addition this commit added the ability to configure initial rules for the filter components in the meta program. This was done here in preparation for the router unification which uses these changes to enforce the desired webserver filtering namely (port 80, tcp traffic and only accessible via internal network).

### DIFF
--- a/examples/firewall/specs.py
+++ b/examples/firewall/specs.py
@@ -1,0 +1,584 @@
+# Copyright 2025, UNSW SPDX-License-Identifier: BSD-2-Clause
+from dataclasses import dataclass, field
+from typing import List, Optional, Dict, Any, Tuple
+from sdfgen import SystemDescription, Sddf, DeviceTree
+from config_structs import (
+    RegionResource,
+    DeviceRegionResource,
+    FwConnectionResource,
+    FwArpConnection,
+    FwDataConnectionResource,
+)
+
+ProtectionDomain = SystemDescription.ProtectionDomain
+MemoryRegion = SystemDescription.MemoryRegion
+Map = SystemDescription.Map
+Channel = SystemDescription.Channel
+
+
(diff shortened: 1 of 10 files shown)


include/lions/firewall/common.h
```diff
@@ -45,15 +45,12 @@ stored big-endian, so mask byte order must be swapped for subnet match. */
 
 /* Firewall ID number used by components to identify which interface they
 are connected to */
-#define FW_EXTERNAL_INTERFACE_ID 0
-#define FW_INTERNAL_INTERFACE_ID 1
+#define FW_EXTERNAL_INTERFACE_ID 1
+#define FW_INTERNAL_INTERFACE_ID 0
```
cazb2, line 49: This is to match the setup in the meta program, but the constants themselves will be removed when the routing component gets unified since there is no need to hard-code support for two interfaces.

examples/firewall/specs.py
```diff
@@ -0,0 +1,584 @@
+# Copyright 2025, UNSW SPDX-License-Identifier: BSD-2-Clause
+from dataclasses import dataclass, field
+from typing import List, Optional, Dict, Any, Tuple
+from sdfgen import SystemDescription, Sddf, DeviceTree
+from config_structs import (
```
cazb2, line 5: I think it would make a lot of sense to define these structs in the python code and generate the required config.h C file from corresponding python classes

Issue #230, opened 2025-10-15
Firewall automated testing


Courtney3141:
As the firewall and its network environment can now be run in docker, automated testing scripts which test all functionality can now be created. This would make it easier for developers to identify unanticipated breaking changes to unrelated functionalities.

PR #274, opened 2026-02-16
Firewall: Implement automated testing

adugina:
# Firewall: Implement automated testing

Contributes to the completion of #230

This pull request introduces a testing script that can be executed from within the [Docker firewall environment](https://lionsos.org/docs/examples/firewall/docker/). The tests primarily call external programs (for example, `ping` and `netcat`), so they have been implemented in POSIX shell with assistance from the [shUnit2](https://github.com/kward/shunit2) testing framework. The existing `testing.sh` script was used as a reference.

One new file has been introduced: the `autotest.sh` script; and the `Dockerfile` has been modified to add additional test dependencies: `shunit2` (for setup, tear down and temporary file handling), `curl` (to perform requests to the firewall's web API) and `jq` (to generate a JSON POST request from a firewall rule's parameters).

Typically, a test calls an external tool, and then its output is captured and interpreted.

## Demo

[Automated testing demo](https://github.com/user-attachments/assets/d418dfc2-cd23-4c5e-9c67-11884a6672be)

### Tests

For the ICMP module, there are tests that
- ping hosts on opposing sides of the firewall,
- ping unreachable hosts to confirm that an unreachable message is received, and
- ping the firewall directly, from both the internal and external networks.

For both TCP and UDP, there are tests that
- transmit traffic from an internal host to an external and vice versa.

There is also a test that
- applies a firewall rule, tests its application, and then removes it.

(Tests exist to ping the firewall directly, but they are currently skipped because this feature has not been implemented yet.)

### Miscellaneous

In addition,
- temporary files are cleaned up automatically,
- tests can be run selectively; that is, one or several tests can be run, rather than the entire test suite, and
- toggles are available to
	- use user-provided or randomly generated data for data transmission tests,
	- display network interface and routing information on start up, and
	- produce firewall logs (preserving coloured output) on test failure or after test suite completion.

## Documentation

Documentation is provided within the test script, and [another pull request](https://github.com/au-ts/lionsos.org/pull/25) contains user documentation to be added to the LionsOS website.

### DIFF
--- a/examples/firewall/docker/scripts/autotest.sh
+++ b/examples/firewall/docker/scripts/autotest.sh
@@ -19,8 +19,9 @@
 # `UDP_PORT` and `TCP_PORT` for `EXT_HOST_IP` and `INT_HOST_IP`.
 #
 # The shUnit2 framework is used for setup, teardown and temporary file handling.
-# For further information on shUnit2, see its documentation available at:
-# https://github.com/kward/shunit2
+# For further information on shUnit2 and its execution behaviour, please refer
+# to the quickstart guide available at:
+# https://github.com/kward/shunit2?tab=readme-ov-file#-quickstart
 
 #
 # Constants
@@ -518,6 +519,7 @@ test_rule_application_and_removal() {
 # shUnit
 #
 
-# Load shUnit2
-#
+# Once shUnit2 has been sourced, it will find all functions that begin with the
+# name `test` and add them to a list to be executed. The source statement should
+# be the last line in the file.
 . shunit2


Courtney3141:
If this script incorporates `testing.sh`, do you think it's worth removing the "raw" script and instead using the single test API you provide? I would prefer that, unless you see a problem with it?

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -0,0 +1,523 @@
...
+
+    # Verify that the data was transmitted correctly
+    if ! diff "${SENT}" "${RECEIVED}" > /dev/null 2>&1; then
+        fail "${ERROR_RULE_STILL_APPLIED}"
+        print_log
+    fi
+}
+
+#
+# shUnit
```
Courtney3141, line 519: Are these comments required for `shUnit`?

Courtney3141:
Happy to merge upon you answering the above. Thanks for this!

adugina:
I'll start with this comment first.

>Actually, I think it may be worth keeping it.
>
>I am just in the process of testing your code, from the code review everything looks great. Only issue so far is that building >the docker image takes way longer. May we can provide an option in the dockerfile to opt in or out of autotest? What do >you think?

I just spent some time investigating the slow build issues that you're experiencing. I don't think it's a consequence of the new test dependencies. On my machine, it takes about 4 minutes to reach the point where it installs `shunit2`, `curl` and `jq`. It then only takes a few seconds to install these new dependencies, and another 10–20 seconds to finish building the image. I tested building the image with and without the new dependencies and they both took a similar length (around 4–5 minutes) – installs earlier in the build process seem to be consuming a lot of time (the `apt-get update` seems incredibly slow today).

Regarding the comment that preceded the one above:

> If this script incorporates `testing.sh`, do you think it's worth removing the "raw" script and instead using the single test API you provide? I would prefer that, unless you see a problem with it?

Most of the commands and comments in the `testing.sh` script are incorporated in to the testing documentation on the LionsOS website, so I don't see any issue with removing the raw script. Most of the contents would be preserved there and provide necessary context to understand what the automated testing script is doing under-the-hood.

It looks like setting up a dummy endpoint is not implemented or documented elsewhere currently though. What's implemented in each script and what's documented on the LionsOS website is listed below. 

|                                   | `testing.sh` | `autotest.sh` | LionsOS website |
| --------------------------------- | :----------: | :-----------: | :-------------: |
| Executing in namespaces           |      ✔️      |      ✔️       |       ✔️        |
| Test ICMP                         |      ✔️      |      ✔️       |       ✔️        |
| Test ICMP destination unreachable |      ✔️      |      ✔️       |       ✔️        |
| Test TCP                          |      ✔️      |      ✔️       |       ✔️        |
| Test UDP                          |      ✔️      |      ✔️       |       ✔️        |
| Monitor traffic with `tcpdump`    |      ✔️      |       ❌       |       ✔️        |
| Display route information         |      ✔️      |      ✔️       |        ❌        |
| Display network interface details |      ✔️      |      ✔️       |        ❌        |
| Setup a dummy endpoint            |      ✔️      |       ❌       |        ❌        |

I guess these would be the next steps.

- Remove `testing.sh`
- Remove mention of `testing.sh` within `autotest.sh` and on the LionsOS website
- (Optionally) mention on the LionsOS website how to display routes and network interface details
- (Optionally) mention on the LionsOS website how to setup a dummy endpoint (I'm not familiar with how they're used in testing currently though)

examples/firewall/docker/scripts/autotest.sh
adugina: No, the comments are optional. We only require that `. shunit2` is the last line of the script.

adugina: Additional context has been added in 473725454b44f505dfad0f4ec4d3cc3c436f10af.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -518,6 +519,7 @@ test_rule_application_and_removal() {
 # shUnit
 #
 
-# Load shUnit2
-#
+# Once shUnit2 has been sourced, it will find all functions that begin with the
```
Courtney3141, line 522: This is great thanks!

PR #276, opened 2026-02-18
Firewall: Apply SDDF style formatting

Courtney3141:


components/micropython/modfirewall.c
```diff
@@ -3,32 +3,32 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-#include <stdint.h>
```
Courtney3141, line 6: Let's not enforce a certain order on our includes - this is not done in sDDF.

cazb2: fixed

PR #291, opened 2026-03-24
Pass MAKEFLAGS down to submake

wom-bat:
also make system file depend on Makefile variables, so it gets rebuilt if they change.

### DIFF
--- a/examples/kitty/Makefile
+++ b/examples/kitty/Makefile
@@ -29,7 +29,7 @@ REPORT_FILE := $(BUILD_DIR)/report.txt
 all: ${IMAGE_FILE}
 
 qemu ${IMAGE_FILE} ${REPORT_FILE} clean clobber: ${BUILD_DIR}/Makefile FORCE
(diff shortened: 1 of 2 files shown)


dumsum:
Two questions/comments since I've been thinking about this a bit.


1. Is the explicit MAKEFLAGS passing needed? I believe this should already be passed automatically to recursive make calls: https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html
This sDDF commit message also made reference to removing explicit MAKEFLAGS passing for the same reason: https://github.com/au-ts/sddf/commit/e2f87a91218d537d7ffa5f5f6a415b64504a7fd2

2. If you first run `make NFS_SERVER=oldval`, and then run `make NFS_SERVER=newval`, the second invocation should now correctly use `newval`. But `build/Makefile` is not regenerated, so a later `make -C build` will still fall back to the snapshotted `oldval`. Is that the intended behaviour? (Edit: sorry just realised this was already observed by @midnightveil above)

PR #302, opened 2026-03-26
Firewall: Handle IPV4 Header Options

cazb2:
Modified code to handle the fact that the IPV4 header length is not always equal to the minimum when optional fields are included.

ICMP_FILTER.C: use transport_layer_offset rather than assuming hdr offsets.
ICMP_MODULE.C: use ipv4_header_length rather than assuming minimum ipv4 hdr len & use offset(data) rather than double counding the payload.
ROUTING.C: use transport_layer_offset rather than assuming minimum hdrs.
ICMP.H: changed ICMP_ECHO_LEN constant to not double count payload len. Updated offset calculations to use transport_layer_offset rather than assuming minimum hdr length. Use offset(icmp_echo_t, data) to retrieve offset to the data part of the icmp_echo_t struct.

### DIFF
--- a/examples/firewall/filters/icmp_filter.c
+++ b/examples/firewall/filters/icmp_filter.c
@@ -116,7 +116,7 @@ static void filter(void)
                 break;
             }
             case FILTER_ACT_REJECT: {
-                icmp_hdr_t *icmp_hdr = (icmp_hdr_t *)(pkt_vaddr + ICMP_HDR_OFFSET);
+                icmp_hdr_t *icmp_hdr = (icmp_hdr_t *)(pkt_vaddr + transport_layer_offset(ip_hdr));
                 uint8_t icmp_type = icmp_hdr->type;
                 /* An ICMP error message shouldn't be sent upon reception of another ICMP error message */
                 if (!icmp_is_error_message(icmp_type)) {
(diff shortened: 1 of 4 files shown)


Courtney3141 (approved):

Courtney3141:
While this is an improvement, in that you stop us from using the wrong IP header length in a number of places, I still think we are incorrectly constructing ICMP packets that contain oversized IP headers. While we are now actually inserting the payload into the `data` field, when we would have been inserting oversized header bytes, I am pretty sure the RFC is fairly explicit that we should be sending back the *entire* IPv4 header - and we are currently truncating it to the minimums size: https://www.rfc-editor.org/rfc/rfc792.html

To properly fix this, I think we either need to include a maximally expanded version of the IPv4 header in our codebase, and use this version for anything involving the ICMP module, or we could just add an array with the extra bytes the header may need to the `icmp_req_t` struct which can be used if the IP header is larger than the minimum (I think I prefer this).

In either case, having a variable sized IP header messes up all of our ICMP packet definitions in icmp.h that include the header. At least it only effects the entry after the header, which is data. I propose when we eventually fix this, we should probably remove the data fields from the packet definitions, rename the last entry to `ip_hdr_and_data`and calculate how many bytes to copy each time. Shouldn't be too difficult to do.

PR #306, opened 2026-03-30
i2c: micropython fixes

midnightveil:
Ready to go.


components/micropython/machine_i2c.c
```diff
@@ -87,59 +102,119 @@ static i2c_err_t mp_i2c_dispatch(machine_i2c_obj_t *self, uint16_t addr, uint8_t
...
-            // FIXME: not an assert.
-            assert(i2c_bus_release(i2c_config.virt.id, addr));
-            return ret;
-        }
-        num_acks += ret;
-    }
+    debug_printf("machine_i2c_transfer: done (err: %d)\n", err);
 
-    // FIXME: not an assert.
+    // always release the bus regardless of the return (FIXME: not-assert)
```
omeh-a, line 190: Is it worth releasing the bus at all? Our current security model assumes that once clients claim an address it will never be released, largely due to the fact that it's basically just a race to claim stuff. Once the new sdfgen version exists we will be statically assigning buses, which means that there will be no way to do this at all.

omeh-a (commented):
Looks good to me. Much needed dose of sanity. Only one oddity re: releasing the bus, because I don't think we really need to do this.

omeh-a (approved):

components/micropython/machine_i2c.c
midnightveil: (discussed in person) This is how it was before, it's possible that later we should remove this.

PR #311, opened 2026-04-01
Firewall ICMP autotests

DaiDaiDai123:
Added autotests for :
1. ICMP timeout in both directions
2. ICMP reject for UDP and ICMP interface in both directions
3. ICMP ping disable response

Changed the ping response for all interface to be defaultly enabled.


include/lions/firewall/icmp.h
```diff
@@ -118,7 +118,8 @@ typedef struct __attribute__((__packed__)) icmp_echo {
 } icmp_echo_t;
 
 /* Total length of ICMP echo request/reply packet with maximum payload */
-#define ICMP_ECHO_LEN (ICMP_COMMON_HDR_LEN + sizeof(icmp_echo_t) + FW_ICMP_ECHO_PAYLOAD_LEN)
+#define ICMP_ECHO_LEN (ICMP_COMMON_HDR_LEN + ICMP_PAYLOAD_OFFSET + FW_ICMP_ECHO_PAYLOAD_LEN)
```
Courtney3141, line 121: The original `#define ICMP_ECHO_LEN` here is definitely wrong - I have a feeling this is a remnant from back before you included the data field inside the `typedef for icmp_echo_t`.
However, your new definition is not correct either - `ICMP_PAYLOAD_OFFSET` gives the byte offset of the ICMP packet payload within the entire Ethernet frame, or in other words the first byte after the ICMP header. `ICMP_ECHO_LEN` is supposed to give the total (maximum) number of bytes in an ICMP echo request/reply, including the header. So the payload offset is irrelevant here.

This should now be:
`#define ICMP_ECHO_LEN (ICMP_COMMON_HDR_LEN + sizeof(icmp_echo_t))`

include/lions/firewall/icmp.h
```diff
@@ -118,7 +118,8 @@ typedef struct __attribute__((__packed__)) icmp_echo {
...
+#define ICMP_ECHO_HDR_LEN (sizeof(uint16_t) + sizeof(uint16_t))
```
Courtney3141, line 122: The reason I was using `sizeof(icmp_echo_t) - FW_ICMP_ECHO_PAYLOAD_LEN` for calculating payload length was to deduce how many bytes of the `data` field the echo request packet was using - I.e. how many bytes of the packet are left after accounting for the `id` and `seq` fields.

I am happy to introduce a shortcut macro for this, however I would rather define as
```
(sizeof(icmp_echo_t) - FW_ICMP_ECHO_PAYLOAD_LEN)
```
rather than `(sizeof(uint16_t) + sizeof(uint16_t))`. I understand they will give the same result, however defining it like I suggested gives more intuition for what we are using the macro for, and is hopefully less confusing than the sum of two integer sizes.

Additionally I think the name you have chosen is a misnomer, since there is no such thing as an *echo header* - the echo packet is encapsulated within an ICMP packet, the header of which is given by the type `icmp_hdr_t`. So I would prefer a name like `ICMP_ECHO_NO_DATA_LEN`.

include/lions/firewall/icmp.h
Courtney3141: Luckily it seems we weren't using this field anywhere...

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+    ip netns exec int \
+        ping -c "${COUNT}" -w "${TIMEOUT}" "${FW_INT_IP}" > "${RECEIVED}" 2>&1
+
+    if grep -Eq --ignore-case "${REGEX_REACHABLE}" "${RECEIVED}"; then
+        fail "${ERROR_UNEXPECTED_ECHO_RESPONSE}"
+        print_file 'Ping output' "${RECEIVED}"
+        print_log
+    fi
+
+    # Restore ping responsiveness so subsequent tests are unaffected.
```
Courtney3141, line 401: Rather than assuming that ping responsiveness was originally turned on, could you instead:
- Use the API to get what it was originally set to
- If it is set to `on`, turn it `off`
- Run the test
- If it was originally `on`, turn it back `on`

This also lets us test the `getPing` endpoint.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+        "http://${FW_INT_IP}/api/ping/1/1")
+
+    if echo "${response}" | grep -q '"error"'; then
+        fail "${ERROR_FAILED_ENABLE_PING}"
+        print_log
+    fi
+}
+
+test_icmp_ping_firewall_disabled_from_external_network() {
+    # Disable ping responsiveness on the external interface.
```
Courtney3141, line 414: Please make the suggested change for this test too.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -35,13 +35,19 @@ ERROR_NO_ECHO_RESPONSE='Did not receive echo response'
...
 ERROR_DATA_WAS_NOT_DROPPED='Firewall traffic was not dropped'
 ERROR_FAILED_TO_APPLY_RULE='Failed to apply firewall rule'
 ERROR_FAILED_TO_REMOVE_RULE='Failed to remove firewall rule'
 ERROR_RULE_STILL_APPLIED='Firewall rule is still applied'
+ERROR_ICMP_REJECT_NOT_APPLIED='ICMP reject rule did not block ping traffic'
+ERROR_UDP_REJECT_DID_NOT_TERMINATE='UDP reject did not terminate sender traffic'
+ERROR_UDP_REJECT_UNEXPECTED_OUTPUT='UDP reject produced unexpected sender output'
+ERROR_UNEXPECTED_ECHO_RESPONSE='Received echo response when ping should be disabled'
 INFO_SKIPPING_TEST='Skipping (feature not implemented yet)'
+ERROR_TIMEOUT='Did not receive timeout'
```
Courtney3141, line 50:
```suggestion
ERROR_TIMEOUT='Did not receive time to live exceeded message'
```

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -35,13 +35,19 @@ ERROR_NO_ECHO_RESPONSE='Did not receive echo response'
...
 ERROR_TRANSMIT_FAILED='Failed to transmit data'
 ERROR_DATA_INCORRECT='The received data is different to what was sent'
 ERROR_DATA_WAS_NOT_DROPPED='Firewall traffic was not dropped'
 ERROR_FAILED_TO_APPLY_RULE='Failed to apply firewall rule'
 ERROR_FAILED_TO_REMOVE_RULE='Failed to remove firewall rule'
 ERROR_RULE_STILL_APPLIED='Firewall rule is still applied'
+ERROR_ICMP_REJECT_NOT_APPLIED='ICMP reject rule did not block ping traffic'
+ERROR_UDP_REJECT_DID_NOT_TERMINATE='UDP reject did not terminate sender traffic'
+ERROR_UDP_REJECT_UNEXPECTED_OUTPUT='UDP reject produced unexpected sender output'
+ERROR_UNEXPECTED_ECHO_RESPONSE='Received echo response when ping should be disabled'
```
Courtney3141, line 48:
```suggestion
ERROR_UNEXPECTED_ECHO_RESPONSE='Received echo response when ping responsiveness is disabled`
```

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -50,14 +56,17 @@ FONT_RESET=$(printf '\033[0m')
 REGEX_REACHABLE='[1-9][0-9]* received'
 REGEX_HOST_UNREACHABLE='Destination Host Unreachable'
 REGEX_NET_UNREACHABLE='Destination Net Unreachable'
+REGEX_TIMEOUT='Time to live exceeded'
 
 TEMPLATE_SRC='$src_ip, $src_port, $src_subnet'
 TEMPLATE_DEST='$dest_ip, $dest_port, $dest_subnet'
 TEMPLATE_ACTION='$interface, $action'
```
Courtney3141, line 63: I know you haven't changed this, but your PR made me notice it. I don't think it makes sense to call the `TEMPLATE_ACTION` field this, since it includes the interface... Could you please split them up? Also, can you please rename `TEMPLATE_JSON` to `TEMPLATE_RULE_JSON`

```sh
TEMPLATE_INTERFACE='$interface'
TEMPLATE_ACTION='$action'
TEMPLATE_JSON="{ ${TEMPLATE_SRC}, ${TEMPLATE_DEST}, ${TEMPLATE_INTERFACE}, ${TEMPLATE_ACTION} }"
```

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+        --request 'DELETE' \
+        "http://${FW_INT_IP}/api/rules/icmp/${rule_id}/${FIREWALL_INTERNAL_INTERFACE}")
+
+    if echo "${response}" | grep -q '"error"'; then
+        fail "${ERROR_FAILED_TO_REMOVE_RULE}"
+        print_log
+        return
+    fi
+
+    # Ping should not succeed while the reject rule is active.
```
Courtney3141, line 514: You have the right idea here - putting the check *after* the rule has been removed means that the rule will be taken down even if the test fails (if the rule stuck around in the fail case, then it would cause future tests to fail).

However, since this check *does not* actually return in the case of failure, I think it should be done immediately after the ping has been sent - see `test_rule_application_and_removal` as an example.

In particular, the way it is structured now means that if the removal of the rule failed for some reason, then we would not know whether the ping was blocked or not (since we return before this check in that case).

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+        --request 'DELETE' \
+        "http://${FW_INT_IP}/api/rules/icmp/${rule_id}/${FIREWALL_EXTERNAL_INTERFACE}")
+
+    if echo "${response}" | grep -q '"error"'; then
+        fail "${ERROR_FAILED_TO_REMOVE_RULE}"
+        print_log
+        return
+    fi
+
+    # Ping should not succeed while the reject rule is active.
```
Courtney3141, line 566: My comment about the ordering above applies here too.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+        "http://${FW_INT_IP}/api/rules/icmp/${rule_id}/${FIREWALL_INTERNAL_INTERFACE}")
+
+    if echo "${response}" | grep -q '"error"'; then
+        fail "${ERROR_FAILED_TO_REMOVE_RULE}"
+        print_log
+        return
+    fi
+
+    # Ping should not succeed while the reject rule is active.
+    if grep -Eq --ignore-case "${REGEX_REACHABLE}" "${RECEIVED}"; then
```
Courtney3141, line 515: This just checks that no ping responses were received - however, I don't think this distinguishes from the case where the firewall just blocks the ping. I think you should specifically add a regex for checking whether a `Destination Port Unreachable` packet was received, similar to the `REGEX_HOST_UNREACHABLE` regex.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+        "http://${FW_INT_IP}/api/rules/icmp/${rule_id}/${FIREWALL_EXTERNAL_INTERFACE}")
+
+    if echo "${response}" | grep -q '"error"'; then
+        fail "${ERROR_FAILED_TO_REMOVE_RULE}"
+        print_log
+        return
+    fi
+
+    # Ping should not succeed while the reject rule is active.
+    if grep -Eq --ignore-case "${REGEX_REACHABLE}" "${RECEIVED}"; then
```
Courtney3141, line 567: My comment about the more specific regex applies here too.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -35,13 +35,19 @@ ERROR_NO_ECHO_RESPONSE='Did not receive echo response'
...
 ERROR_NET_UNREACHABLE='Did not receive destination net unreachable'
 ERROR_FAILED_ENABLE_PING='Could not enable ping response'
+ERROR_FAILED_DISABLE_PING='Could not disable ping response'
 ERROR_TRANSMIT_FAILED='Failed to transmit data'
 ERROR_DATA_INCORRECT='The received data is different to what was sent'
 ERROR_DATA_WAS_NOT_DROPPED='Firewall traffic was not dropped'
 ERROR_FAILED_TO_APPLY_RULE='Failed to apply firewall rule'
 ERROR_FAILED_TO_REMOVE_RULE='Failed to remove firewall rule'
 ERROR_RULE_STILL_APPLIED='Firewall rule is still applied'
+ERROR_ICMP_REJECT_NOT_APPLIED='ICMP reject rule did not block ping traffic'
```
Courtney3141, line 45:
```suggestion
ERROR_ICMP_REJECT_NOT_APPLIED='ICMP reject rule did not block ping traffic or send a destination port unreachable message'
```

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+        --request 'DELETE' \
+        "http://${FW_INT_IP}/api/rules/udp/${rule_id}/${FIREWALL_EXTERNAL_INTERFACE}")
+
+    if echo "${response}" | grep -q '"error"'; then
+        fail "${ERROR_FAILED_TO_REMOVE_RULE}"
+        print_log
+        return
+    fi
+
+    # In this environment, reject terminates sender traffic without returning
```
Courtney3141, line 621: Move this check to before the rule is removed

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+        fail "${ERROR_FAILED_TO_APPLY_RULE}"
+        return
+    fi
+
+    # Send UDP traffic and capture sender-side output where ICMP reject details
+    # are reported.
+    ip netns exec ext \
+        nc -u -w "${LONG_TIMEOUT}" "${INT_HOST_IP}" "${UDP_PORT}" \
+        < "${SENT}" > /dev/null 2> "${RECEIVED}"
+    exit_code=$?
```
Courtney3141, line 607: I see that you record this value, however it is not used. After looking at your success check below, I am not convinced that your test actually differentiates between the three cases (the traffic was allowed, the traffic was simply blocked, or the traffic was blocked *and* a port unreachable packet was received in response).

I did some quick research into netcat, and ran some tests, and it seems that if you use the `-z` flag, this stands for "(Zero-I/O): Scans for open ports without sending data, often used for port scanning". In this mode, if a port unreachable packet is received this causes `nc`'s error code to change from 0 to 1. In particular, if the traffic is simply blocked, it seems that `nc` is unaware of this, and it will print a "Connection to 172.16.2.200 655 port [udp/tinc] succeeded!" message and return a code of 0.


So I propose that you use the `-z` option, and then to check for successful application of the reject rule you can simply check for a non-zero exit code.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -35,13 +35,19 @@ ERROR_NO_ECHO_RESPONSE='Did not receive echo response'
...
 ERROR_FAILED_ENABLE_PING='Could not enable ping response'
+ERROR_FAILED_DISABLE_PING='Could not disable ping response'
 ERROR_TRANSMIT_FAILED='Failed to transmit data'
 ERROR_DATA_INCORRECT='The received data is different to what was sent'
 ERROR_DATA_WAS_NOT_DROPPED='Firewall traffic was not dropped'
 ERROR_FAILED_TO_APPLY_RULE='Failed to apply firewall rule'
 ERROR_FAILED_TO_REMOVE_RULE='Failed to remove firewall rule'
 ERROR_RULE_STILL_APPLIED='Firewall rule is still applied'
+ERROR_ICMP_REJECT_NOT_APPLIED='ICMP reject rule did not block ping traffic'
+ERROR_UDP_REJECT_DID_NOT_TERMINATE='UDP reject did not terminate sender traffic'
```
Courtney3141, line 46: This error is not used.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -35,13 +35,19 @@ ERROR_NO_ECHO_RESPONSE='Did not receive echo response'
...
+ERROR_UDP_REJECT_UNEXPECTED_OUTPUT='UDP reject produced unexpected sender output'
```
Courtney3141, line 47:
```suggestion
ERROR_UDP_REJECT_UNEXPECTED_OUTPUT='UDP reject rule did not send a destination port unreachable message`
```

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+    if [ -z "${rule_id}" ]; then
+        fail "${ERROR_FAILED_TO_APPLY_RULE}"
+        return
+    fi
+
+    # Send UDP traffic and capture sender-side output.
+    ip netns exec int \
+        nc -u -w "${LONG_TIMEOUT}" "${EXT_HOST_IP}" "${UDP_PORT}" \
+        < "${SENT}" > /dev/null 2> "${RECEIVED}"
+    exit_code=$?
```
Courtney3141, line 661: Same comment above applies here.

examples/firewall/docker/scripts/autotest.sh
```diff
@@ -366,6 +376,310 @@ test_icmp_ping_firewall_from_external_network() {
...
+
+    if echo "${response}" | grep -q '"error"'; then
+        fail "${ERROR_FAILED_TO_REMOVE_RULE}"
+        print_log
+        return
+    fi
+
+    # In this environment, reject terminates sender traffic without returning
+    # a textual error message to userspace.
+    if ! diff /dev/null "${RECEIVED}" > /dev/null 2>&1; then
```
Courtney3141, line 677: Move this before rule removal.

examples/firewall/docker/scripts/autotest.sh
Courtney3141: Could you also do the same for the `test_icmp_ping_firewall_from_*_network` tests? In face, I would be happy for to combine the ping disabled + ping enabled tests into one for each interface - something like `test_icmp_ping_firewall_from_*_network`:
- Use the API to get what it was originally set to
- If it is set to on, turn it off
- Check no responses to ping
- Turn it on
- Check for responses to ping
- If it was originally off, turn it back off

Courtney3141 (changes requested):

Issue #280, opened 2026-03-10
Add a firewall autotest for ensuring broadcast traffic is dropped


Courtney3141:
There were still a number of issues with your implementation, although it was an improvement. I rebased the code on the current version of main - this is why I squashed your commit history. Then I made some fixes myself in a separate commit. Happy to merge this now, once the CI checks pass.

I need you to write an autotest for this though in a different PR. I am happy to keep it to one simple test - that broadcast MAC address traffic is indeed blocked. To test this, you can get one host to send a broadcast packet, and ensure that the other side doesn't receive it. I did this using the `socat` library:

```
echo "Hello" | ip netns exec ext socat - UDP-DATAGRAM:255.255.255.255:555,broadcast
```

This library is not currently installed in our container, but you can add this to the package list in the same PR where you update the `autotest.sh` script - this was a recent addition to the example by Alex. Also, if you can add a very brief description of your test to the lionsos website, lionsos.org. The repo for this website is here: https://github.com/au-ts/lionsos.org/, make a PR to this repo, in particular you should only need to modify this page: https://lionsos.org/docs/examples/firewall/testing/, and only a brief description.

_Originally posted by @Courtney3141 in https://github.com/au-ts/lionsos/issues/256#issuecomment-4027930337_

PR #312, opened 2026-04-11
Firewall broadcast test

ChronoComplexity:


Courtney3141 (approved):

Courtney3141:
This PR closes issue #280 .

I have added a few more tests:
- broadcasting from both directions
- Sending traffic to both subnet broadcast addresses 

I also realised that you can send broadcast with net cat, so I switched back to netcat to avoid having to include another package. If you could please update the LionsOS website as well, we can finish all the broadcast handling work.

PR #316, opened 2026-05-12
Perform further libc initialisation

larrabyte:
`musllibc` expects programs to start executing via `_start` and so its start routine contains calls to internal initialisation routines which among other things, allocates memory for thread-local storage. This PR adds these internal routines to the tail of our custom `init_libc` so that we also initialise `musllibc` as expected.

### DIFF
--- a/lib/libc/posix/posix.c
+++ b/lib/libc/posix/posix.c
@@ -119,6 +119,11 @@ static long sys_getgid(va_list ap) {
     return 501;
 }
 
+static long sys_set_tid_address(va_list ap) {
+    (void)ap;
+    return 0;
+}
+
 // FIXME: this is deliberately insecure for now
 static long sys_getrandom(va_list ap) {
     void *buf = va_arg(ap, void *);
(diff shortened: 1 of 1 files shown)


dumsum:
One thing to check: `__init_tp` makes a `set_tid_address` syscall, but this is unimplemented (should return `-ENOSYS` I think?). Will that cause any issues?

larrabyte:
As far as I can tell `set_tid_address` is used to notify other threads that a thread has terminated, so I think I can probably provide a stub along the lines of what is already done for `getpid`/`getuid`/`getgid` (return 0 or some default reasonable value) given that we're running in a single-threaded environment.

Forgot to extract it out of my existing work, so I'll commit an implementation now.

dumsum:
> (return 0 or some default reasonable value)

I can see `tid` is used across a bunch of `pthread_mutex` functions. I'm sure most or all of this isn't relevant for us but see for example
https://github.com/au-ts/musllibc/blob/81842bc43d0b49bbcca03ff5f506c6df8e2bddd9/src/thread/pthread_mutex_trylock.c#L10-L23

For a `tid` of 0, the `own == tid` check will always pass even for unlocked.

With that in mind, might be safer to go with a default of 1 instead of 0. `getpid` should probably also be changed to return 1 in this case so that `gettid` is correct ("In a single-threaded process, the thread ID is equal to the process ID")

larrabyte:
Makes sense, modified `getpid` and `set_tid_address` to return 1.

dumsum (approved):

PR #317, opened 2026-05-13
kitty: fix network hang on qemu by removing ethernet device from guest DTS

dumsum:


Ivan-Velickovic (changes requested):
Can you provide more context on why this is suddenly needed? I'm confused.

dumsum:
after discussion, I'll work on a broader mmio fixup and open a new PR

PR #321, opened 2026-05-20
libc: clients provide heap memory

dumsum:
Instead of a fixed static array in the library itself, the `libc_init` API has now changed to allow library users to provide their own region for malloc etc.

### DIFF
--- a/lib/libc/posix/mem.c
+++ b/lib/libc/posix/mem.c
@@ -3,9 +3,10 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+#include <assert.h>
 #include <errno.h>
-#include <stdalign.h>
 #include <stdarg.h>
+#include <stddef.h>
 #include <stdint.h>
 #include <sys/mman.h>
 #include <sys/types.h>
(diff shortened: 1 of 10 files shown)


lib/libc/posix/mem.c
```diff
@@ -3,28 +3,20 @@
...
 #define PAGE_SIZE 0x1000
 
-/*
- * Statically allocated morecore area.
- *
- * This is rather terrible, but is the simplest option without a
- * huge amount of infrastructure.
- */
-#define MORECORE_AREA_BYTE_SIZE 0x100000
-static alignas(8) char morecore_area[MORECORE_AREA_BYTE_SIZE];
```
Ivan-Velickovic, line 23: Do we want to check that the user-supplied pointer is 8-byte aligned?

dumsum: As far as I can tell this doesn't appear to be a correctness issue, but unaligned pointers could be wasteful. musl does some alignment itself with `brk`. `mmap` page aligns anyway.

If anything we could enforce page alignment? Seems reasonable in this situation.

PR #323, opened 2026-05-24
Firewall Implementation: NAT

Hazingoo:
Implements #188

This PR adds NAT (Network Address Translation) support to the LionsOS firewall, enabling internal network devices to share a single public IP address when communicating with external networks.

### Design

The NAT implementation is split across two layers:

**NAT Module (`nat_module.h` / `nat_module.c`)**
A translation engine used by both the RX and TX virtualisers. Each virtualiser holds two module instances, one for TCP, one for UDP. The module handles:
- **SNAT** (outbound): rewrites the internal source IP:port to the public SNAT IP and an assigned ephemeral port
- **DNAT** (inbound/returning): uses the ephemeral port as a direct table index to reverse the translation back to the original internal IP:port
- Checksum recalculation for both the IP header and TCP/UDP transport header after any modification

**Port Table**
A fixed pre allocated array shared between the RX and TX virtualisers for the same interface and protocol. Each slot stores the original internal IP and port, a validity flag, a last used timestamp, and a linked list pointer for the free pool. The port number itself is used as the array index (port - base_port), giving O(1) lookup with no hashing.

**Webserver State**
A shared memory region mapped into the RX virtualiser, TX virtualiser, and webserver simultaneously. The webserver writes the SNAT IP (configured via the web UI) and timeout value; the NAT modules read them directly. A magic number prevents double initialisation when multiple virtualisers start up.

## Features to be added later

- **Timer integration**: `now` is currently hardcoded to `0`
- **Periodic cleanup**: `nat_module_cleanup_expired()` exists but is never called, needs to be implemented.
- **Persistent SNAT IP config**: the configured SNAT IP lives in runtime memory only and is lost on reboot
- **Per protocol timeouts**: TCP and UDP currently share the same 5-second default timeout
- **Fragmented packet handling**: assumes the transport header is present in the first fragment


components/micropython/modfirewall.c
```diff
@@ -509,6 +511,71 @@ static mp_obj_t rule_get_nth(mp_obj_t interface_idx_in, mp_obj_t protocol_in, mp
 
 static MP_DEFINE_CONST_FUN_OBJ_3(rule_get_nth_obj, rule_get_nth);
 
+/* NAT API functions */
+
+/* nat_set_enabled(interface, protocol, enabled) — PPC call to TX and RX virtualizers */
+static mp_obj_t nat_set_enabled(mp_obj_t interface_idx_in, mp_obj_t protocol_in, mp_obj_t enabled_in)
+{
+    uint8_t interface_idx = mp_obj_get_int(interface_idx_in);
```
Courtney3141, line 519: We have now wrapped this error handling in a function, please do the same:

```c
uint8_t interface_idx = mp_obj_get_int(interface_idx_in);
if (!check_interface_index(interface_idx)) {
    return mp_const_none;
}
```

components/micropython/modfirewall.c
```diff
@@ -509,6 +511,71 @@ static mp_obj_t rule_get_nth(mp_obj_t interface_idx_in, mp_obj_t protocol_in, mp
...
+    if (interface_idx >= FW_NUM_INTERFACES) {
+        sddf_dprintf("WEBSERVER|LOG: %s\n", fw_os_err_str[OS_ERR_INVALID_INTERFACE]);
+        mp_raise_OSError(OS_ERR_INVALID_INTERFACE);
+        return mp_const_none;
+    }
+
+    uint8_t protocol = mp_obj_get_int(protocol_in);
```
Courtney3141: You need to check this is a valid protocol:
```c
uint16_t protocol = mp_obj_get_int(protocol_in);
int8_t protocol_match = find_filter_index(interface_idx, protocol);
if (protocol_match == FW_MAX_FILTERS) {
    return mp_const_none;
}
```

include/lions/firewall/config.h
```diff
@@ -23,9 +23,12 @@
 #define FW_MAX_INITIAL_FILTER_RULES 16
 #define FW_MAX_INITIAL_ROUTES 16
 #define FW_MAX_ARP_REQUESTER_CLIENTS 2
+#define FW_MAX_NAT 16
```
Courtney3141: Max NAT what? what is a NAT?

Courtney3141: I see that this is the max number of port tables per interface (although this is not used consistently in the webserver config).

Since for each interface, there is one port table per protocol, I think you can use the value `FW_MAX_FILTERS` instead, which is the maximum number of (protocol) filters per interface. Then you can remove this new macro.

include/lions/firewall/config.h
```diff
@@ -39,6 +40,20 @@ typedef struct fw_data_connection_resource {
     device_region_resource_t data;
 } fw_data_connection_resource_t;
 
+/* NAT interface configuration (used by both RX and TX virtualizers) */
+typedef struct fw_nat_interface_config {
```
Courtney3141: This is not an "interface" config (i.e. a config that is allocated per interface). Each interface will have *multiple* of these. Thus I would rename this to `fw_nat_port_table_config` -> as it contains all the initial information needed to use a port mapping table.

include/lions/firewall/config.h
```diff
@@ -39,6 +40,20 @@ typedef struct fw_data_connection_resource {
...
+    uint16_t base_port;
+    uint16_t ports_capacity;
+    region_resource_t port_table;
+    uint32_t ip;
```
Courtney3141: Since the IP address is fixed (source IP --> IP of the firewall on the transmission interface), this field should not be here (as it does not vary per port mapping). Since the virtualiser's do not have the interface IP address in their config by default, you can add it as a field in the `fw_net_virt_[rt]x_config_t` structs.

include/lions/firewall/config.h
```diff
@@ -39,6 +40,20 @@ typedef struct fw_data_connection_resource {
...
+} fw_nat_interface_config_t;
+
+typedef struct fw_virt_rx_nat_config {
```
Courtney3141: Firstly, this name does not make sense since the type is used by both the rx and tx virtualisers. More importantly though, I don't think the `protocol` and `enabled` fields need to be in a separate struct - I think you should move them into the above struct.

include/lions/firewall/config.h
```diff
@@ -47,6 +62,10 @@ typedef struct fw_net_virt_tx_config {
     uint8_t num_data_regions;
     fw_data_connection_resource_t free_clients[FW_MAX_FW_CLIENTS];
     uint8_t num_free_clients;
+    bool nat_enabled;
```
Courtney3141: Since we not enable NAT on a *protocol* level for an interface, there is no point having this additional configuration information. To figure out whether NAT is enabled for a given packet, the virtualisers will need to check the corresponding protocol port table config.

include/lions/firewall/config.h
```diff
@@ -39,6 +40,20 @@ typedef struct fw_data_connection_resource {
...
+typedef struct fw_nat_interface_config {
+    uint16_t base_port;
+    uint16_t ports_capacity;
+    region_resource_t port_table;
+    uint32_t ip;
+} fw_nat_interface_config_t;
+
+typedef struct fw_virt_rx_nat_config {
+    fw_nat_interface_config_t interface_config;
+    uint8_t protocol;
```
Courtney3141, line 47: This should be a uint16_t

include/lions/firewall/config.h
```diff
@@ -47,6 +62,10 @@ typedef struct fw_net_virt_tx_config {
     uint8_t num_data_regions;
     fw_data_connection_resource_t free_clients[FW_MAX_FW_CLIENTS];
     uint8_t num_free_clients;
+    bool nat_enabled;
+    region_resource_t webserver_state;
```
Courtney3141: What is this used for? Rather than mapping in an *additional* region to share with the webserver, I propose that the webserver can just have read only access to all the port table regions. Then the port table regions can contain the following struct:

```c
bool enabled; // Since this changes at run-time
uint16_t base_port; // This *may* change at run-time in the future
uint16_t ports_size; // Optional - if mappings are contiguous, this tracks the last valid mapping
port_map_t port_table[];
```

I.e. the shared region contains the metadata, followed by a variable length port mapping array. This is how we do it for the routing table (see `fw_routing_table_t`) and the filter rules tables (see `fw_rule_table_t`)

include/lions/firewall/config.h
```diff
@@ -47,6 +62,10 @@ typedef struct fw_net_virt_tx_config {
...
+    fw_virt_rx_nat_config_t nat_configs[FW_MAX_NAT];
```
Courtney3141: Use `FW_MAX_FILTERS` for array size here.

include/lions/firewall/config.h
```diff
@@ -59,6 +78,11 @@ typedef struct fw_net_virt_rx_config {
     uint16_t active_client_subtypes[SDDF_NET_MAX_CLIENTS];
     fw_connection_resource_t free_clients[FW_MAX_FW_CLIENTS];
     uint8_t num_free_clients;
+    bool nat_enabled;
```
Courtney3141: Above applies to this boolean and `webserver_state` as well.

include/lions/firewall/config.h
```diff
@@ -59,6 +78,11 @@ typedef struct fw_net_virt_rx_config {
...
+    region_resource_t webserver_state;
+    device_region_resource_t nat_dma_region;
```
Courtney3141, line 78: I'm confused.. what is this? The Rx virtualiser only needs to read the contents of packets located in one Rx DMA region right? Why have you re-defined it here? It is passed to the Rx virtualiser in it's *sddf network virtualiser* config struct `net_virt_rx_config_t`.

Courtney3141: Oh, I guess it is so it has write permissions for the packet! I see :-) Could you maybe add a one line C comment explaining this please.

include/lions/firewall/config.h
```diff
@@ -163,11 +187,18 @@ typedef struct fw_webserver_interface_config {
     fw_connection_resource_t rx_free;
 } fw_webserver_interface_config_t;
 
+typedef struct fw_webserver_nat_protocol_config {
```
Courtney3141: Rather than defining a new type here, I propose that you just re-use the struct defined above, that I suggested renaming to `fw_nat_port_table_config`. Then:

1. Include an array of size `FW_MAX_FILTER` in the `fw_webserver_interface_config_t` (not the `fw_webserver_config_t`). This is because the port tables are defined *per interface*, not globally.
2. The webserver can then use the `protocol`, `capacity`, and `port_table` fields to identify each port table and the corresponding metadata like `enabled` (which will be stored *inside* the table region).

E.g. request to see if NAT is enabled on interface 1, protocol UDP:

```c
for (i = 0; i < web_config.interfaces[1].num_ports; i++) {
    if (web_config.interfaces[1].nat_configs[i].protocol == UDP) {
        port_table *table = (port_table *)web_config.interfaces[1].nat_configs[i].port_table.vaddr;
        return table->enabled;
```
3. Although the struct also contains the `enabled` field, this will be ignored by the webserver (since it changes at run-time, the webserver can read it from shared memory). If you plan to allow `base_port` to be updated at run time, the same applies. If it is fixed, the webserver can also read it directly from the config (useful if you want to display it on the UI).

components/micropython/modfirewall.c
```diff
@@ -509,6 +511,71 @@ static mp_obj_t rule_get_nth(mp_obj_t interface_idx_in, mp_obj_t protocol_in, mp
...
+    bool enabled = mp_obj_is_true(enabled_in);
+
+    for (uint8_t i = 0; i < fw_config.num_nat_state; i++) {
+        if (fw_config.nat_state[i].protocol == protocol &&
+            fw_config.nat_state[i].interface == interface_idx) {
+            microkit_mr_set(NAT_SET_ENABLED_ARG_ENABLED, (seL4_Word)enabled);
+            (void)microkit_ppcall(fw_config.nat_state[i].tx_ch,
```
Courtney3141: Hmm, I am not so happy about calling two virtualisers. For one, it introduces a race condition. I propose that you just call *one*, and you designate responsibility to the "master" virtualiser for updating the `enabled` field in the table. Since the Tx virtualiser writes to the table much more frequently, I propose the master being the Tx virt. Then the Rx virtualiser just needs to check the table each time it receives a packet.

NOTE: Let's return to this later, as it can be needlessly expensive for a non-NAT'd Rx virtualiser to have to check the table each packet. I think we could store a cached value in the Rx virt, so if the cached variable says NAT is disabled, you don't bother checking the table. Then, the Tx virt could PPC (or signal?) the Rx virt when it changes.... We can discuss later.

Additionally, you need to check for errors here, i.e.

```c
fw_os_err_t os_err = filter_err_to_os_err(microkit_mr_get(FILTER_RET_ERR));
```

components/micropython/modfirewall.c
```diff
@@ -509,6 +511,71 @@ static mp_obj_t rule_get_nth(mp_obj_t interface_idx_in, mp_obj_t protocol_in, mp
...
+    if (interface_idx >= FW_NUM_INTERFACES) {
+        return mp_const_false;
+    }
+
+    uint8_t protocol = mp_obj_get_int(protocol_in);
+
+    for (uint8_t i = 0; i < fw_config.num_nat_state; i++) {
+        if (fw_config.nat_state[i].protocol == protocol &&
+            fw_config.nat_state[i].interface == interface_idx) {
+            microkit_msginfo reply = microkit_ppcall(fw_config.nat_state[i].tx_ch,
```
Courtney3141: As discussed in the config file, this can be read directly from the table without a ppc

components/micropython/modfirewall.c
```diff
@@ -509,6 +511,71 @@ static mp_obj_t rule_get_nth(mp_obj_t interface_idx_in, mp_obj_t protocol_in, mp
...
+                                                     microkit_msginfo_new(NAT_GET_ENABLED, 0));
+            fw_nat_err_t err = (fw_nat_err_t)microkit_mr_get(NAT_RET_ERR);
+            (void)reply;
```
Courtney3141: Not sure what this line of code does, but regardless please handle errors in the same way as the other functions in this file. The wrappers I wrote ensure the error gets propagated to Python in a useful way.

examples/firewall/net_components/firewall_network_components.mk
```diff
@@ -33,11 +33,17 @@ ${FIREWALL_NETWORK_COMPONENT_OBJ}: |firewall_network/net_components $(SDDF_LIBC_
 ${FIREWALL_NETWORK_COMPONENT_OBJ}: ${CHECK_FIREWALL_NETWORK_FLAGS_MD5}
 ${FIREWALL_NETWORK_COMPONENT_OBJ}: CFLAGS+=${CFLAGS_FIREWALL_NETWORK}
 
-firewall_network/net_components/firewall_network_virt_%.o: ${SDDF}/firewall_network/net_components/virt_%.c |firewall_network/net_components
+firewall_network/net_components/firewall_network_virt_%.o: ${FIREWALL_NETWORK_COMPONENTS_DIR}/firewall_network_virt_%.c |firewall_network/net_components
```
Courtney3141, line 36: Good catch, but please create a similar rule for the nat object file.

examples/firewall/net_components/firewall_network_components.mk
```diff
@@ -33,11 +33,17 @@ ${FIREWALL_NETWORK_COMPONENT_OBJ}: |firewall_network/net_components $(SDDF_LIBC_
...
 	${CC} ${CFLAGS} -c -o $@ $<
 
-%.elf: firewall_network/net_components/%.o |firewall_network/net_components
-	${LD} ${LDFLAGS} -o $@ $< ${LIBS}
+firewall_network/net_components/nat_module.o: ${FIREWALL_NETWORK_COMPONENTS_DIR}/nat_module.c |firewall_network/net_components
+	${CC} ${CFLAGS} -c -o $@ $<
+
+firewall_network_virt_rx.elf: firewall_network/net_components/firewall_network_virt_rx.o firewall_network/net_components/nat_module.o |firewall_network/net_components
```
Courtney3141: You should be able to combine these two rules using a wildcard, as was previously done. You are just adding one extra dependency. Look into the definitions of the automatic variables ($@ $< etc) to see if you can figure it out.

examples/firewall/net_components/firewall_network_virt_rx.c
```diff
@@ -5,18 +5,23 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+#include <string.h>
```
Courtney3141, line 7: Do we need this?

examples/firewall/net_components/firewall_network_virt_rx.c
```diff
@@ -26,6 +31,10 @@ net_queue_handle_t rx_queue_clients[SDDF_NET_MAX_CLIENTS];
 
 fw_queue_t fw_free_clients[FW_MAX_FW_CLIENTS];
 
+/* NAT modules for TCP and UDP */
+static nat_module_t nat_tcp_module;
```
Courtney3141: These should be arrays....

`nat_module_t nat_modules[FW_MAX_FILTER]`

Also, I don't know why you declared them static?

examples/firewall/net_components/firewall_network_virt_rx.c
```diff
@@ -68,8 +77,21 @@ static void rx_return(void)
             int err = net_dequeue_active(&rx_queue_drv, &buffer);
             assert(!err);
 
-            buffer.io_or_offset = buffer.io_or_offset - config.data.io_addr;
-            uintptr_t buffer_vaddr = buffer.io_or_offset + (uintptr_t)config.data.region.vaddr;
+            uintptr_t data_vaddr;
```
Courtney3141, line 87: It might be cleaner to always use the read write permission region (shouldn't do any harm if you are not modifying the packets)

examples/firewall/net_components/firewall_network_virt_rx.c
```diff
@@ -86,6 +108,41 @@ static void rx_return(void)
             //
             // [1]: https://developer.arm.com/documentation/ddi0595/2021-06/AArch64-Instructions/DC-IVAC--Data-or-unified-Cache-line-Invalidate-by-VA-to-PoC
             cache_clean_and_invalidate(buffer_vaddr, buffer_vaddr + buffer.len);
+
+            /* Apply NAT translation if enabled */
+            if (fw_config.nat_enabled)
```
Courtney3141: NAT is performed per protocol, so you need to extract the protocol and check the corresponding table...

examples/firewall/net_components/firewall_network_virt_rx.c
```diff
@@ -86,6 +108,41 @@ static void rx_return(void)
...
+            {
+                uint16_t ethtype = htons(((eth_hdr_t *)buffer_vaddr)->ethtype);
+                if (ethtype == ETH_TYPE_IP)
+                {
+                    ipv4_hdr_t *ip_hdr = (ipv4_hdr_t *)(buffer_vaddr + IPV4_HDR_OFFSET);
+                    /* RX path receives packets from external network, so they are inbound */
+                    bool is_inbound = true;
```
Courtney3141: This variable is always true, I don't think you need to define it, you can just pass true directly.

examples/firewall/net_components/firewall_network_virt_rx.c
```diff
@@ -86,6 +108,41 @@ static void rx_return(void)
...
+
+                    int nat_result = NAT_SUCCESS;
+                    if (ip_hdr->protocol == IPV4_PROTO_TCP)
```
Courtney3141: This will be a for loop through your port tables, searching for the protocol number. If the protocol number is found, check the table to see if NAT is currently enabled.

Courtney3141: NOTE: We may want to have a special case, where if *no* NAT is enabled, we skip the search altogether (i.e. cache the value of `enabled_0 || enabled_1 || ... || enabled_n`)

examples/firewall/net_components/firewall_network_virt_rx.c
```diff
@@ -86,6 +108,41 @@ static void rx_return(void)
...
+                    {
+                        nat_result = nat_module_translate(&nat_tcp_module, buffer_vaddr, &buffer, is_inbound);
+                    }
+                    else if (ip_hdr->protocol == IPV4_PROTO_UDP)
+                    {
+                        nat_result = nat_module_translate(&nat_udp_module, buffer_vaddr, &buffer, is_inbound);
+                    }
+
+                    /* Drop packet if NAT translation fails */
+                    if (nat_result != NAT_SUCCESS)
```
Courtney3141: Perhaps we want to configure what we do if the protocol does not match with a port table entry

Courtney3141: Additionally, we may want to have a way to add a port table wildcard protocol - i.e., you receive a packet whose protocol does not have a table, it can be placed in the wildcard table.

Courtney3141 (commented):
I spent a couple of hours looking through, I have identified a number of things to work on.

1. I would start by reading through my comments on config.h, which motivate my other comments.
2. The main issue I have identified so far with the implementation is that you half-half whether you implement NAT on a protocol level or an interface level. My comments assume that we should implement on a protocol level, which I don't think is a bad idea. But a lot of your code assumes that NAT is either enabled or disabled on an *interface* - you need to fix that.
3. If NAT *is* enabled, the virtualisers are required to search through their port tables, which reside in shared memory. This can actually be expensive if it is done on a per-packet basis. Thus, I think we need to implement some optimisations for the case where NAT is disabled. I described how the enabled boolean should reside in the port table itself - this still stands, but you may want to introduce a local cache for each virtualiser of whether *any* protocol NAT is enabled. This allows the virtualiser to skip port table searching in the totally disabled case. The virt can then rely on a ppc from the webserver (or potentially other virt) to indicate whether the cache needs to be updated. (This cache could actually be a uint8_t indicating how many protocols it is enabled on). Feel free to discuss this more with me next week if you want.
4. I deleted some old code from `ui_server.py` that seemed to be lingering in your repository, however your endpoints may have depended on it. You will need to update them if so (this was old code assuming that there are only 2 interfaces, so you can't assume this any more.)

If you are confused about any of my comments so far, let me know and we can discuss online or in-person.

examples/firewall/net_components/nat_module.c
```diff
@@ -0,0 +1,291 @@
+/*
+ * Copyright 2026, UNSW
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#include "microkit.h"
```
Courtney3141, line 5: Is this really required? Doesn't seem like it.

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+#include <sddf/util/printf.h>
+#include <sddf/network/queue.h>
+#include <sddf/timer/protocol.h>
+#include <lions/firewall/config.h>
+#include <lions/firewall/common.h>
+
+/**
+ * NAT module success/error codes
+ */
+#define NAT_SUCCESS 0
```
Courtney3141: In other places in the code, we use enums for this, could you please match the surrounding style?

```c
typedef enum {
    /* no error */
    ROUTING_ERR_OKAY = 0,
    /* data structure is full */
    ROUTING_ERR_FULL,
    /* duplicate entry exists */
    ROUTING_ERR_DUPLICATE,
    /* entry clashes with existing entry */
    ROUTING_ERR_CLASH,
    /* node ID does not point to a valid node */
    ROUTING_ERR_INVALID_ID,
    /* route is invalid */
    ROUTING_ERR_INVALID_ROUTE,
} fw_routing_err_t;
```

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+#define NAT_FAILURE -1
+#define NAT_PORT_EXHAUSTED -2
+#define NAT_INVALID_PACKET -3
+
+/* NAT timeout interval in nanoseconds */
```
Courtney3141, line 28: Is this the amount of time that passes before a port mapping becomes invalid? If so, 5 seconds seems a little short.

include/lions/firewall/nat_module.h
Courtney3141: Also, best to add descriptions, what is `NAT_FAILURE`?

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+ * The SNAT IP is static (from build-time config); only enable/disable and timeout are runtime.
+ */
+typedef struct fw_nat_webserver_state
+{
+    uint32_t magic; /* Magic number for initialization check */
+    /* Timeout in nanoseconds */
+    uint64_t timeout;
+} fw_nat_webserver_state_t;
+
+/**
```
Courtney3141: If you have a .c file, you may as well define *all* your functions in the C file rather than some in the header...

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+static inline fw_nat_port_mapping_t *fw_nat_translate_destination(fw_nat_interface_config_t interfaces[],
+                                                                  uint8_t num_interfaces,
+                                                                  uint32_t dst_ip,
+                                                                  uint16_t dst_port,
+                                                                  uint64_t now)
+{
+    /* Since dst_port is used as an index here it must be in host byte order */
+    dst_port = htons(dst_port);
+
+    for (uint16_t i = 0; i < num_interfaces; i++)
```
Courtney3141: I don't understand why you are looking through *all* interfaces here, shouldn't we only look through the port mappings of the interface it was received on ?

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+        {
+            fw_nat_port_table_t *port_table = (fw_nat_port_table_t *)interfaces[i].port_table.vaddr;
+
+            if ((dst_port >= interfaces[i].base_port) && (dst_port < interfaces[i].base_port + port_table->largest_index) && port_table->mappings[dst_port - interfaces[i].base_port].is_valid)
+            {
+                return &port_table->mappings[dst_port - interfaces[i].base_port];
+            }
+        }
+    }
+
```
Courtney3141: How does this function not reference protocol number at all? If there is a table for each protocol...

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+                                                                  uint32_t dst_ip,
+                                                                  uint16_t dst_port,
+                                                                  uint64_t now)
+{
+    /* Since dst_port is used as an index here it must be in host byte order */
+    dst_port = htons(dst_port);
+
+    for (uint16_t i = 0; i < num_interfaces; i++)
+    {
+        if (dst_ip == interfaces[i].ip)
```
Courtney3141: There should be a unique IP per interface, so if this block fails then the function should shortcut early to an error. However as I said, I would have thought this function should be checking all *protocol* port mapping tables on a single interface.

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+ * @return Ephemeral port in network byte order, or 0 if no port available
+ */
+static inline uint16_t fw_nat_find_ephemeral_port(fw_nat_interface_config_t config,
+                                                  fw_nat_port_table_t *ports,
+                                                  uint32_t src_ip,
+                                                  uint16_t src_port,
+                                                  uint64_t now)
+{
+    /* Search for an existing mapping */
+    for (uint16_t i = 0; i < ports->size; i++)
```
Courtney3141, line 108: the entries are not necessarily contiguous right? So you can't use size here.

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+        ports->free_head = mapping->next_free;
+
+        mapping->src_port = src_port;
+        mapping->src_ip = src_ip;
+        mapping->last_used_ts = now;
+        mapping->is_valid = true;
+        return htons(config.base_port + (mapping - ports->mappings));
+    }
+
+    if (ports->size >= config.ports_capacity)
```
Courtney3141: This should really be an  `else` block to the previous statement...

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+    {
+        return 0; /* Ephemeral ports pool is full */
+    }
+
+    /* Initialize a new entry in the NAT table and assign its ephemeral port */
+    ports->size++;
```
Courtney3141: What is this code doing here? I think this was a remnant of when Hamzah didn't initialise the free list at init time, but we discussed that he *should* do that.... I thought he fixed it.

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * @param config NAT config for this interface
+ * @param ports Ephemeral port table for this interface
+ * @param src_ip Source IP in network byte order
+ * @param src_port Source port in network byte order
+ * @param now Current timestamp in nanoseconds
+ *
+ * @return Ephemeral port in network byte order, or 0 if no port available
+ */
+static inline uint16_t fw_nat_find_ephemeral_port(fw_nat_interface_config_t config,
```
Courtney3141: Probably makes more sense for this function to return an error code, and the result port as pointer.

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+static inline void fw_nat_free_ephemeral_port(fw_nat_interface_config_t config,
+                                              fw_nat_port_table_t *ports,
+                                              uint16_t port)
+{
+    fw_nat_port_mapping_t *mapping = &ports->mappings[port - config.base_port];
+
+    mapping->src_ip = 0;
+    mapping->src_port = 0;
+    mapping->next_free = ports->free_head;
+    mapping->is_valid = false;
```
Courtney3141, line 153: this is all you need to do, no point setting other entries to 0

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+static inline void fw_nat_free_expired_mappings(fw_nat_interface_config_t config,
+                                                fw_nat_port_table_t *ports,
+                                                uint64_t timeout,
+                                                uint64_t now)
+{
+    for (uint16_t i = 0; i < ports->largest_index; i++)
+    {
+        fw_nat_port_mapping_t *mapping = &ports->mappings[i];
+
+        if (mapping->is_valid && now > timeout && mapping->last_used_ts <= now - timeout)
```
Courtney3141, line 179:
```suggestion
        if (mapping->is_valid && mapping->last_used_ts + timeout <= now)
```

This is again more feedback I left on Hamzah's original PR...

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+    uint8_t interface;
+
+    /* Protocol (IPPROTO_TCP or IPPROTO_UDP) */
+    uint8_t protocol;
+
+    /* Port table reference (shared memory) — contains nat_enabled flag */
+    fw_nat_port_table_t *port_table;
+
+    /* Interface configuration (shared memory) */
+    fw_nat_interface_config_t *interface_config;
```
Courtney3141: i don’t think this is actually in shared memory?

include/lions/firewall/nat_module.h
```diff
@@ -0,0 +1,339 @@
...
+
+    /* Byte offsets for protocol-specific header parsing */
+    size_t src_port_off; /* Offset to source port in transport header */
+    size_t dst_port_off; /* Offset to destination port in transport header */
+    size_t check_off;    /* Offset to checksum in transport header */
+
+    /* Whether to recalculate checksum */
```
Courtney3141, line 213:
```suggestion
    /* Whether to recalculate transport layer checksum */
```

examples/firewall/net_components/nat_module.c
```diff
@@ -0,0 +1,291 @@
...
+                    mapping->last_used_ts = now;
+
+                    *dst_port = mapping->src_port;
+                    ip_hdr->dst_ip = mapping->src_ip;
+                    ip_hdr->check = 0;
+
+                    checksum_dirty = true;
+                    nat->dnat_hits++;
+                    dnat_done = true;
+                }
```
Courtney3141, line 138: no else? I wish I had known you re-implemented these functions in the header, as I have already reviewed the other functions... You should have removed them.

examples/firewall/net_components/nat_module.c
```diff
@@ -0,0 +1,291 @@
...
+    return NAT_SUCCESS;
+}
+
+/**
+ * Translate a packet using NAT
+ */
+int nat_module_translate(nat_module_t *nat,
+                         uintptr_t pkt_vaddr,
+                         net_buff_desc_t *buffer,
+                         bool do_dnat)
```
Courtney3141, line 61: I really think these should be different functions to be honest. (Translation on the way in/out)

examples/firewall/net_components/nat_module.c
```diff
@@ -0,0 +1,291 @@
...
+                    nat->dnat_hits++;
+                    dnat_done = true;
+                }
+            }
+        }
+    }
+
+    /* SNAT: outbound traffic leaving the internal network; rewrite source to iface_ip:ephemeral_port
+     * and record the mapping so returning traffic can be DNAT-ed back. */
+    if (!dnat_done && nat->port_table->nat_enabled && ip_hdr->dst_ip != nat->interface_config->ip)
```
Courtney3141: `ip_hdr->dst_ip != nat->interface_config->ip` this should never happen - the firewall won't send out a packet for itself.

Courtney3141: I also feel like the nat enabled check should be done once at the beginning of the function (or even before calling it).

examples/firewall/net_components/nat_module.c
```diff
@@ -0,0 +1,291 @@
...
+    {
+        *check = 0;
+        *check = calculate_transport_checksum(
+            transport_hdr,
+            htons(ip_hdr->tot_len) - ipv4_header_length(ip_hdr),
+            nat->protocol,
+            ip_hdr->src_ip,
+            ip_hdr->dst_ip);
+
+        ip_hdr->check = 0;
```
Courtney3141, line 191: actually, this always has to be done. The reason the check enabled exists is about whether the transport layer protocol has a checksum. UDP and TCP do, ICMP doesn't. But regardless of transport layer, IP will always need to be recalculated.

examples/firewall/net_components/nat_module.c
```diff
@@ -0,0 +1,291 @@
...
+        uint8_t ihl_bytes = ip_hdr->ihl * 4;
+        ip_hdr->check = fw_internet_checksum(ip_hdr, ihl_bytes);
```
Courtney3141, line 193:
```suggestion
        ip_hdr->check = fw_internet_checksum(ip_hdr, ipv4_header_length(ip_hdr))```

File: examples/firewall/net_components/firewall_network_virt_tx.c [hunk 46]
```diff
@@ -23,6 +29,10 @@ net_queue_handle_t tx_queue_clients[SDDF_NET_MAX_CLIENTS];
 fw_queue_t fw_free_clients[FW_MAX_FW_CLIENTS];
 fw_queue_t fw_active_clients[FW_MAX_FW_CLIENTS];
 
+/* NAT modules for TCP and UDP */
+static nat_module_t nat_tcp_module;
```
Courtney3141: These should really be arrays

PR #325, opened 2026-05-28
posix/file.c: fix stat for root directory

dumsum:
Background: FatFS rejects "/" with FR_INVALID_NAME, so `fstat_int` had a stub to return success for a stat call returning this error (FS_STATUS_INVALID_NAME from the FS Server). However, `statbuf` was left unpopulated, so callers potentially read uninitialised memory. 

https://github.com/au-ts/lionsos/pull/321 exposed this latent bug: it broke `wasm_test` because WAMR's WASI preopen registration seemingly couldn't see "/" as a directory (I suspect this just happened to work by chance before the morecore change).

Fix is to populate a minimal stat for "/".

### DIFF
--- a/lib/libc/posix/file.c
+++ b/lib/libc/posix/file.c
@@ -267,6 +267,14 @@ static int file_dup3(int oldfd, int newfd) {
 }
 
 static int fstat_int(const char *path, struct stat *statbuf) {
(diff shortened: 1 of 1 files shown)


midnightveil (approved):
This seems fine to me, but I don't know much about FATFS.

PR #330, opened 2026-07-08
included full IPV4 header in ICMP error packets

### LINKED ISSUE #315: Include the entire IPv4 header in our ICMP packets generated by the ICMP module
While this is an improvement, in that you stop us from using the wrong IP header length in a number of places, I still think we are incorrectly constructing ICMP packets that contain oversized IP headers. While we are now actually inserting the payload into the `data` field, when we would have been inserting oversized header bytes, I am pretty sure the RFC is fairly explicit that we should be sending back the *entire* IPv4 header - and we are currently truncating it to the minimums size: https://www.rfc-editor.org/rfc/rfc792.html

To properly fix this, I think we either need to include a maximally expanded version of the IPv4 header in our codebase, and use this version for anything involving the ICMP module, or we could just add an array with the extra bytes the header may need to the `icmp_req_t` struct which can be used if the IP header is larger than the minimum (I think I prefer this).

In either case, having a variable sized IP header messes up all of our ICMP packet definitions in icmp.h that include the header. At least it only effects the entry after the header, which is data. I propose when we eventually fix this, we should probably remove the data fields from the packet definitions, rename the last entry to `ip_hdr_and_data`and calculate how many bytes to copy each time. Shouldn't be too difficult to do.

_Originally posted by @Courtney3141 in https://github.com/au-ts/lionsos/issues/302#issuecomment-4420471803_

yumnazaheed:
This PR updates the ICMP module so that the ICMP packets it generates now include the full IPv4 header. This improves packet completeness and works to ensurethe generated ICMP messages better match the expected network behavior described in issue #315.

Fixes #315

### DIFF
--- a/examples/firewall/icmp/icmp_module.c
+++ b/examples/firewall/icmp/icmp_module.c
@@ -86,7 +86,7 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
     uint8_t orig_ip_hdr_len = ipv4_header_length(&req->ip_hdr);
     /* Calculate how many bytes of the original packet's data to include  */
     uint16_t orig_pkt_data_len = ntohs(req->ip_hdr.tot_len) - orig_ip_hdr_len;
-    uint16_t to_copy = MIN(FW_ICMP_SRC_DATA_LEN, orig_pkt_data_len);
+    uint16_t data_to_copy = MIN(FW_ICMP_SRC_DATA_LEN, orig_pkt_data_len);
 
     /* Handle each ICMP type separately */
     switch (req->type) {
@@ -118,7 +118,7 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
         * including the unused field, length field, next-hop MTU field,
         * the original IPv4 header, and the accompanying payload.
         */
-        uint16_t icmp_dest_payload_len = 4 + orig_ip_hdr_len + to_copy;
+        uint16_t icmp_dest_payload_len = ICMP_DEST_LEN_NO_IP + orig_ip_hdr_len + data_to_copy;
         ip_hdr->tot_len = htons(IPV4_HDR_LEN_MIN + ICMP_COMMON_HDR_LEN + icmp_dest_payload_len);
 
         /* Construct ICMP destination unreachable packet */
@@ -137,7 +137,7 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
         }
 
         /* Copy first bytes of data if applicable */
-        memcpy(&icmp_dest->ip_data[orig_ip_hdr_len], req->dest.data, to_copy);
+        memcpy(&icmp_dest->ip_data[orig_ip_hdr_len], req->dest.data, data_to_copy);
         break;
     case ICMP_TTL_EXCEED:
         /* Destination is original packet's source */
@@ -146,7 +146,7 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
         /* Calculate the ICMP Time Exceeded packet length,
         * including the unused field, original IPv4 header, and payload.
         */
-        uint16_t icmp_time_exceeded_payload_len = 4 + orig_ip_hdr_len + to_copy;
+        uint16_t icmp_time_exceeded_payload_len = ICMP_TIME_EXCEEDED_LEN_NO_IP + orig_ip_hdr_len + data_to_copy;
         ip_hdr->tot_len = htons(IPV4_HDR_LEN_MIN + ICMP_COMMON_HDR_LEN + icmp_time_exceeded_payload_len);
 
         /* Construct ICMP time exceeded packet */
@@ -163,7 +163,7 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
         }
 
         /* Copy first bytes of data if applicable */
-        memcpy(&icmp_time_exceeded->ip_data[orig_ip_hdr_len], req->time_exceeded.data, to_copy);
+        memcpy(&icmp_time_exceeded->ip_data[orig_ip_hdr_len], req->time_exceeded.data, data_to_copy);
         break;
     case ICMP_REDIRECT_MSG:
         /* Destination is original packet's source */
@@ -172,7 +172,7 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
         /* Calculate the ICMP redirect packet length,
         * including the gateway IP address, original IPv4 header, and payload.
         */
-        uint16_t icmp_redirect_payload_len = 4 + orig_ip_hdr_len + to_copy;
+        uint16_t icmp_redirect_payload_len = ICMP_REDIRECT_LEN_NO_IP + orig_ip_hdr_len + data_to_copy;
         ip_hdr->tot_len = htons(IPV4_HDR_LEN_MIN + ICMP_COMMON_HDR_LEN + icmp_redirect_payload_len);
 
         /* Construct ICMP redirect packet */
@@ -189,7 +189,7 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
         }
 
         /* Copy first bytes of data if applicable */
-        memcpy(&icmp_redirect->ip_data[orig_ip_hdr_len], req->redirect.data, to_copy);
+        memcpy(&icmp_redirect->ip_data[orig_ip_hdr_len], req->redirect.data, data_to_copy);
         break;
     default:
         return false;
--- a/examples/firewall/icmp/icmp_module.c
+++ b/examples/firewall/icmp/icmp_module.c
@@ -82,7 +82,11 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
     icmp_hdr->type = req->type;
     icmp_hdr->code = req->code;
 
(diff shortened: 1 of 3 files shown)


include/lions/firewall/icmp.h
```diff
@@ -87,25 +87,21 @@ typedef struct __attribute__((__packed__)) icmp_dest {
     uint8_t len;
     /* optional MTU of the next-hop network if source packet was too large, or 0 */
     uint16_t nexthop_mtu;
-    /* IP header of source packet */
-    ipv4_hdr_t ip_hdr;
-    /* first 8 bytes of data from source packet */
-    uint8_t data[FW_ICMP_SRC_DATA_LEN];
+    /* IP header (up to 60 bytes) followed by data (up to 8 bytes) */
+    uint8_t ip_data[60 + FW_ICMP_SRC_DATA_LEN];
```
Courtney3141: Could you please create a macro for IP header max length rather than use 60 directly? Like

```c
/* Length of IPv4 header with no optional fields */
#define IPV4_HDR_LEN_MIN sizeof(ipv4_hdr_t)
```

in `include/lions/firewall/ip.h`

include/lions/firewall/icmp.h
```diff
@@ -87,25 +87,21 @@ typedef struct __attribute__((__packed__)) icmp_dest {
...
 } icmp_dest_t;
 
-#define ICMP_DEST_LEN (ICMP_COMMON_HDR_LEN + sizeof(icmp_dest_t))
+#define ICMP_DEST_LEN (ICMP_COMMON_HDR_LEN + 4 + 60 + FW_ICMP_SRC_DATA_LEN)
```
Courtney3141: Although this is correct, I don't think this needs to be changed right? Probably best to use `sizeof`

include/lions/firewall/icmp.h
```diff
@@ -87,25 +87,21 @@ typedef struct __attribute__((__packed__)) icmp_dest {
...
-    /* IP header of source packet */
-    ipv4_hdr_t ip_hdr;
-    /* First 8 bytes of data from source packet */
-    uint8_t data[FW_ICMP_SRC_DATA_LEN];
+    /* IP header (up to 60 bytes) followed by data (up to 8 bytes) */
+    uint8_t ip_data[60 + FW_ICMP_SRC_DATA_LEN];
 } icmp_redirect_t;
 
-#define ICMP_REDIRECT_LEN (ICMP_COMMON_HDR_LEN + sizeof(icmp_redirect_t))
+#define ICMP_REDIRECT_LEN (ICMP_COMMON_HDR_LEN + 4 + 60 + FW_ICMP_SRC_DATA_LEN)
```
Courtney3141: Same here, this does not need to be changed.

include/lions/firewall/icmp.h
```diff
@@ -178,6 +172,7 @@ typedef struct icmp_req {
     eth_hdr_t eth_hdr;
     /* header of source IP packet */
     ipv4_hdr_t ip_hdr;
+    uint32_t ip_hdr_extra[31];
```
Courtney3141: This doesn't look right to me, the maximum size is 60 bytes right? You have added 31 * 4 bytes = 124 bytes (too many). Once you have added the macro I suggested, you could define this field instead as:

```c
uint8_t ip_hdr_extra[IPV4_HDR_LEN_MAX - IPV4_HDR_LEN_MIN];
```

include/lions/firewall/icmp.h
```diff
@@ -226,6 +221,9 @@ static inline bool icmp_enqueue_error(fw_queue_t *icmp_queue, uint8_t type, uint
     /* Copy IP header into ICMP request */
     ipv4_hdr_t *ip_hdr = (ipv4_hdr_t *)(pkt_vaddr + IPV4_HDR_OFFSET);
     memcpy(&req.ip_hdr, (void *)ip_hdr, IPV4_HDR_LEN_MIN);
+    if (ipv4_header_length(ip_hdr) > IPV4_HDR_LEN_MIN) {
+        memcpy(&req.ip_hdr_extra, (void *)ip_hdr + IPV4_HDR_LEN_MIN, ipv4_header_length(ip_hdr) - IPV4_HDR_LEN_MIN);
+    }
 
     /* Copy first bytes of data if applicable */
     uint8_t ip_hdr_len = ipv4_header_length(ip_hdr);
```
Courtney3141: Why don't you define this variable *before* your check above then use it rather than call the function 3 times.

include/lions/firewall/icmp.h
Courtney3141: Also, since it is now a variable length (depending on the length of the original header), it's probably worth renaming this (and the other) macro to `ICMP_DEST_MAX_LEN`

examples/firewall/icmp/icmp_module.c
```diff
@@ -82,7 +82,11 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
     icmp_hdr->type = req->type;
     icmp_hdr->code = req->code;
 
-    uint16_t to_copy = MIN(FW_ICMP_SRC_DATA_LEN, ntohs(req->ip_hdr.tot_len) - ipv4_header_length(&req->ip_hdr));
+    /* Calculate the actual length of the original IP header*/
+    uint8_t orig_ip_hdr_len = ipv4_header_length(&req->ip_hdr);
+    /* Calculate how many bytes of the original packet's data to include  */
+    uint16_t orig_pkt_data_len = ntohs(req->ip_hdr.tot_len) - orig_ip_hdr_len;
+    uint16_t to_copy = MIN(FW_ICMP_SRC_DATA_LEN, orig_pkt_data_len);
```
Courtney3141: Can we rename this to `data_to_copy` please? Just to avoid confusion.

examples/firewall/icmp/icmp_module.c
```diff
@@ -110,8 +114,12 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
         /* Destination is original packet's source */
         ip_hdr->dst_ip = req->ip_hdr.src_ip;
 
-        /* Total length of ICMP destination unreachable IP packet */
-        ip_hdr->tot_len = htons(IPV4_HDR_LEN_MIN + ICMP_DEST_LEN);
+        /* Calculate the ICMP Destination Unreachable packet length,
+        * including the unused field, length field, next-hop MTU field,
+        * the original IPv4 header, and the accompanying payload.
+        */
+        uint16_t icmp_dest_payload_len = 4 + orig_ip_hdr_len + to_copy;
```
Courtney3141: I'm not a huge fan of having hardcoded lengths like this (4), could you please introduce a new macro (perhaps in icmp.h) for this value? Something like

```c
#define ICMP_DEST_LEN_NO_IP (sizeof(icmp_dest_t) - (60 + FW_ICMP_SRC_DATA_LEN))
```

include/lions/firewall/icmp.h
Courtney3141: Actually, after having a look this macro is never used, so let's just remove it :-)

include/lions/firewall/icmp.h
Courtney3141: (let's remove this too)

include/lions/firewall/icmp.h
```diff
@@ -124,13 +120,11 @@ typedef struct __attribute__((__packed__)) icmp_echo {
...
-    /* IP header of source packet */
-    ipv4_hdr_t ip_hdr;
-    /* First 8 bytes of data from source packet */
-    uint8_t data[FW_ICMP_SRC_DATA_LEN];
+    /* IP header (up to 60 bytes) followed by data (up to 8 bytes)*/
+    uint8_t ip_data[60 + FW_ICMP_SRC_DATA_LEN];
 } icmp_time_exceeded_t;
 
-#define ICMP_TIME_EXCEEDED_LEN (ICMP_COMMON_HDR_LEN + sizeof(icmp_time_exceeded_t))
+#define ICMP_TIME_EXCEEDED_LEN (ICMP_COMMON_HDR_LEN + 4 + 60 + FW_ICMP_SRC_DATA_LEN)
```
Courtney3141: Also remove this

examples/firewall/icmp/icmp_module.c
```diff
@@ -121,48 +129,67 @@ static bool process_icmp_request(icmp_req_t *req, bool *transmitted)
...
     case ICMP_TTL_EXCEED:
         /* Destination is original packet's source */
         ip_hdr->dst_ip = req->ip_hdr.src_ip;
 
-        /* Total length of ICMP time exceeded IP packet */
-        ip_hdr->tot_len = htons(IPV4_HDR_LEN_MIN + ICMP_TIME_EXCEEDED_LEN);
+        /* Calculate the ICMP Time Exceeded packet length,
+        * including the unused field, original IPv4 header, and payload.
+        */
+        uint16_t icmp_time_exceeded_payload_len = 4 + orig_ip_hdr_len + to_copy;
```
Courtney3141: Let's define a similar macro for this 4 as well.

Courtney3141 (changes requested):
Looks good to me! Just some minor nitpicks for clarity. Is it passing all the autotests?

Courtney3141 (approved):

PR #331, opened 2026-07-14
Return 0777 for file/directory permissions from FAT driver by default

larrabyte:
The FAT driver does not currently do any permission checks nor expose read/write-only capabilities to its clients, so anyone with a cap to the FAT driver will be able to issue read/write commands. 

This is however not reflected by the driver's implementation of `stat`, as it returns `0755` in the stat structure. Because the structure is initially zeroed with `memset`, the owner UID and GID in the structure will also be 0. The LionsOS libc currently says that any application's user/group ID is 501, so any client checking whether it can write to a given file using `stat` will always think it can't, even though the driver would have allowed a write command.

This is explicitly checked by cFS, so it seems the logical thing to do here is return `0777` for files and directories.

### DIFF
--- a/components/fs/fat/op.c
+++ b/components/fs/fat/op.c
@@ -355,17 +355,20 @@ void handle_stat(void) {
 
     file_stat->size = fileinfo.fsize;
 
-// Now we have only one fat volume, so we can hard code it here
+    // Now we have only one fat volume, so we can hard code it here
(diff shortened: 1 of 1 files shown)


dumsum (approved):
Seems fine, and consistent with other approaches to FAT.

PR #335, opened 2026-08-11
Update sDDF, libVMM and sdfgen dependencies

Courtney3141:


dreamliner787-9:
As of [99d7bb8](https://github.com/au-ts/lionsos/pull/335/commits/99d7bb8d0c61fc11fcacbabb1788d8e7f1b67dd6), the Kitty, webserver and fileio example works on both QEMU and hardware, @dumsum is looking at the wasm and posix examples. @Courtney3141 you might want to retest the firewall in case I broke something

dreamliner787-9:
As for the VMM example, my plan is to let it regress for now. Then I will refresh it later.

dumsum:
posix_test and wasm_test look ok so far

Courtney3141:
The firewall tests are working as well.

midnightveil (approved):

PR #337, opened 2026-08-12
Fixed broken i2c error handling, use exception types instead

### LINKED ISSUE #336: micropython / kitty: unusual i2c behaviour
With the v2 i2c driver, the I2C system is too fast and introduces issues. We have temporarily resolved this for the 2026 Aug release by adding an additional sleep in pn532.py.

This is indicative of a deeper issue however:
* Without the sleep, we crash with a timeout error from the i2c library. This isn't expected! The read_ack_frame code retries reading many times (expecting NACKs or timeouts), but it seems like the error handling for timeouts isn't handled correctly. It should not raise a Python error immediately.
* There theoretically shouldn't be a timeout here at all ... the meson i2c driver is literally incapable of returning a timeout error! This implies that the timeout is coming from somewhere in Micropython (and isn't an I2C error, explaining point 1 above) and we should figure out why. Race condition?

omeh-a:
Closes #336 

Our error handling was borked for i2c. We now use proper exception handling to catch the types that prompt us to retry. This changes error codes to distinguish clearly between bus and sDDF errors, where bus errors are EIO.

### DIFF
--- a/examples/kitty/client/pn532.py
+++ b/examples/kitty/client/pn532.py
@@ -90,7 +90,7 @@ class PN532:
             try:
                 data = self.i2c_bus.readfrom(_PN532_I2C_BUS_ADDRESS, _PN532_ACK_FRAME_SIZE)
             except OSError as e:
-                if e.errno is errno.EIO:
+                if e.errno == errno.EIO:
                     continue    # NACK probably, retry
                 raise e
             if len(data) and data[0] & 1:
@@ -117,7 +117,7 @@ class PN532:
             try:
                 data = self.i2c_bus.readfrom(_PN532_I2C_BUS_ADDRESS, 6)
             except OSError as e:
-                if e.errno is errno.EIO:
+                if e.errno == errno.EIO:
                     continue    # NACK probably, retry
                 raise e
             if (data[0] & 1):
--- a/examples/kitty/client/pn532.py
+++ b/examples/kitty/client/pn532.py
@@ -4,6 +4,7 @@
 # We must ignore these imports since they are not available in a normal Python
 # environment where we do the type checking
 from machine import I2C # type: ignore
+import errno
 from time import sleep_ms # type: ignore
 import sys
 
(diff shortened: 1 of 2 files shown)


examples/kitty/client/pn532.py
```diff
@@ -86,7 +87,12 @@ def _pn532_write_command(self, header: bytearray, body: bytearray, retries: int)
     def _pn532_read_ack_frame(self, retries: int) -> bool:
         attempts = 0
         while (attempts < retries):
-            data = self.i2c_bus.readfrom(_PN532_I2C_BUS_ADDRESS, _PN532_ACK_FRAME_SIZE)
+            try:
+                data = self.i2c_bus.readfrom(_PN532_I2C_BUS_ADDRESS, _PN532_ACK_FRAME_SIZE)
+            except OSError as e:
+                if e.errno is errno.EIO:
```
midnightveil: equals, no guarantee that ints are "is"

examples/kitty/client/pn532.py
```diff
@@ -108,8 +114,12 @@ def _pn532_read_response_length(self, retries: int) -> int:
         attempts = 0
 
         while True:
-            data = self.i2c_bus.readfrom(_PN532_I2C_BUS_ADDRESS, 6)
-
+            try:
+                data = self.i2c_bus.readfrom(_PN532_I2C_BUS_ADDRESS, 6)
+            except OSError as e:
+                if e.errno is errno.EIO:
```
midnightveil: ditto

midnightveil (approved):
Additional context: as far as we can tell, Micropython has no specific error type for an I²C NACK. Most other Micropython NACKs also end up raising a EIO as well, so we follow this convention.

Further additional context: our I²C code prior #306 incorrectly checked for bus errors as `err < 0`, and even when it did return an error, returned a positive number (interpreted by micropython as a length) for error conditions. So previously any NACKs or other timeout errors would be silently ignored, so we were reading nonsense data in the read_ack_frame function for the pn532. Fixing this issue has caused Micropython to start raising exceptions, which we handle here now.

PR #343, opened 2026-08-26
implemented fix to prevent queue overfill

### LINKED ISSUE #292: Firewall queues and network queues may be overfilled
With the merging of #273 it is now possible for queues to become "overfilled", i.e. there are more enque-able buffers in the system than certain queues have the capacity to hold. This does not currently occur in the network subsystem, and since the firewall was modelled off this, it also does not check whether queues are full before enqueuing.

If this were to occur, which it can under high load conditions, either an assert would fail in debug mode, or the queues would become corrupted. This needs to be fixed, and the solution will also require modification of the sDDF network module in sdfgen to allow queue capacity to be specified.

yumnazaheed:
Buffers from multiple interfaces can now converge on certain shared queues, but nothing checked fullness before enqueuing, just assert.
- routing.c: transmit_packet()'s tx_active enqueue and the webserver-routing enqueue now check fw_queue_full() first. if it's full, drop and return the buffer to rx_free.
- firewall_network_virt_tx.c: tx_provide() now checks fullness before forwarding into the driver-shared queue, leaving buffers queued (not dropped) if it's full. 

Fixes #292

### DIFF
--- a/examples/firewall/net_components/firewall_network_virt_tx.c
+++ b/examples/firewall/net_components/firewall_network_virt_tx.c
@@ -23,6 +23,10 @@ net_queue_handle_t tx_queue_clients[SDDF_NET_MAX_CLIENTS];
 fw_queue_t fw_free_clients[FW_MAX_FW_CLIENTS];
 fw_queue_t fw_active_clients[FW_MAX_FW_CLIENTS];
 
+/* Alternates which of the net/fw client groups tx_provide() will serve first, to 
+prevent starvation on tx_queue_drv. */
+bool fw_clients_priority = false;
+
 static int extract_offset_net_client(uintptr_t *phys)
 {
     for (int client = 0; client < config.num_clients; client++) {
@@ -48,13 +52,19 @@ static int extract_offset_fw_client(uintptr_t *phys)
     return -1;
 }
 
-static void tx_provide(void)
+/* sdfgen sizes tx_queue_drv for the net clients only, not the firewall
+clients also feeding it, so fullness must be checked before every enqueue. */
+static void tx_provide_net_clients(bool *enqueued, bool *drv_full)
 {
-    bool enqueued = false;
-    for (int client = 0; client < config.num_clients; client++) {
+    for (int client = 0; client < config.num_clients && !*drv_full; client++) {
         bool reprocess = true;
         while (reprocess) {
             while (!net_queue_empty_active(&tx_queue_clients[client])) {
+                if (net_queue_full_active(&tx_queue_drv)) {
+                    *drv_full = true;
+                    break;
+                }
+
                 net_buff_desc_t buffer;
                 int err = net_dequeue_active(&tx_queue_clients[client], &buffer);
                 assert(!err);
@@ -75,7 +85,11 @@ static void tx_provide(void)
 
                 err = net_enqueue_active(&tx_queue_drv, buffer);
                 assert(!err);
-                enqueued = true;
+                *enqueued = true;
+            }
+
+            if (*drv_full) {
+                break;
             }
 
             net_request_signal_active(&tx_queue_clients[client]);
@@ -87,9 +101,19 @@ static void tx_provide(void)
             }
         }
     }
+}
 
-    for (int client = 0; client < fw_config.num_active_clients; client++) {
+static void tx_provide_fw_clients(bool *enqueued, bool *drv_full)
+{
+    for (int client = 0; client < fw_config.num_active_clients && !*drv_full; client++) {
         while (!fw_queue_empty(&fw_active_clients[client])) {
+            if (net_queue_full_active(&tx_queue_drv)) {
+                /* Leave remaining buffers queued rather than dropping them;
+                picked up on the next tx_provide() once tx_return() frees space. */
+                *drv_full = true;
+                break;
+            }
+
             fw_buff_desc_t buffer;
             int err = fw_dequeue(&fw_active_clients[client], &buffer);
             assert(!err);
@@ -105,9 +129,24 @@ static void tx_provide(void)
             net_buff_desc_t net_buffer = { .io_or_offset = io_addr, .len = buffer.len };
             err = net_enqueue_active(&tx_queue_drv, net_buffer);
             assert(!err);
-            enqueued = true;
+            *enqueued = true;
         }
     }
+}
+
+static void tx_provide(void)
+{
+    bool enqueued = false;
+    bool drv_full = false;
+
+    if (fw_clients_priority) {
+        tx_provide_fw_clients(&enqueued, &drv_full);
+        tx_provide_net_clients(&enqueued, &drv_full);
+    } else {
+        tx_provide_net_clients(&enqueued, &drv_full);
+        tx_provide_fw_clients(&enqueued, &drv_full);
+    }
+    fw_clients_priority = !fw_clients_priority;
 
     if (enqueued && net_require_signal_active(&tx_queue_drv)) {
         net_cancel_signal_active(&tx_queue_drv);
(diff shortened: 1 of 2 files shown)


examples/firewall/net_components/firewall_network_virt_tx.c
```diff
@@ -87,9 +101,19 @@ static void tx_provide(void)
             }
         }
     }
+}
 
-    for (int client = 0; client < fw_config.num_active_clients; client++) {
+static void tx_provide_fw_clients(bool *enqueued, bool *drv_full)
```
Courtney3141: y not use return values?

this still starves individuals.. Let's do round-robin


This doesn't work. You need to make sure not only that you don't overfill the queue, but that you only pass at most the queue's capacity at time, as otherwise this could lead to the *driver* overfilling the queue.

examples/firewall/routing/routing.c
```diff
@@ -79,8 +79,20 @@ static bool enqueue_icmp_unreachable(fw_buff_desc_t buffer, uint32_t next_hop)
     return enqueued;
 }
 
-static void transmit_packet(fw_buff_desc_t buffer, uint8_t *mac_addr, uint8_t out_interface)
+static void drop_buffer(fw_buff_desc_t buffer)
```
Courtney3141: If you call this function each time that `transmit_packet` fails, it might be better to just call the function *within* `transmit_packet` and silently return the buffer if the queue is full.

examples/firewall/routing/routing.c
```diff
@@ -79,8 +79,20 @@ static bool enqueue_icmp_unreachable(fw_buff_desc_t buffer, uint32_t next_hop)
...
+    net_buff_desc_t net_buff = { .io_or_offset = buffer.offset, .len = buffer.len };
+    int err = fw_enqueue(&rx_free[buffer.interface], &net_buff);
+    assert(!err);
+    returned[buffer.interface] = true;
+}
+
+static bool transmit_packet(fw_buff_desc_t buffer, uint8_t *mac_addr, uint8_t out_interface)
+{
+    if (fw_queue_full(&tx_active[out_interface])) {
+        return false;
```
Courtney3141: E.g. add a call to `drop_buffer` here and remove it elsewhere. Additionally, I would rename `drop_buffer` to `drop_packet`.

examples/firewall/routing/routing.c
```diff
@@ -143,7 +156,11 @@ static void process_arp_waiting(uint8_t out_interface)
             /* Substitute the MAC address and send packets out of the NIC */
             pkt_waiting_node_t *node = root;
             for (uint16_t i = 0; i < root->num_children + 1; i++) {
-                transmit_packet(node->buffer, response.mac_addr, out_interface);
+                bool sent = transmit_packet(node->buffer, response.mac_addr, out_interface);
+                if (!sent) {
+                    LOG_FIREWALL("ROUTING", "tx queue for interface %u full, dropping packet\n", out_interface);
```
Courtney3141: You can also add a log for failure in `transmit_packet`

examples/firewall/net_components/firewall_network_virt_tx.c
```diff
@@ -23,6 +23,10 @@ net_queue_handle_t tx_queue_clients[SDDF_NET_MAX_CLIENTS];
 fw_queue_t fw_free_clients[FW_MAX_FW_CLIENTS];
 fw_queue_t fw_active_clients[FW_MAX_FW_CLIENTS];
 
+/* Alternates which of the net/fw client groups tx_provide() will serve first, to
```
Courtney3141: I like this idea, but this can still result in client starvation *within* the firewall clients and net clients, i.e. fw client 0 and net client 0 will always be serviced first.

If you would like to fix the starvation issue, you could possibly round robin through *all* client queues?

examples/firewall/net_components/firewall_network_virt_tx.c
```diff
@@ -48,13 +52,19 @@ static int extract_offset_fw_client(uintptr_t *phys)
...
 }
 
-static void tx_provide(void)
+/* sdfgen sizes tx_queue_drv for the net clients only, not the firewall
+clients also feeding it, so fullness must be checked before every enqueue. */
+static void tx_provide_net_clients(bool *enqueued, bool *drv_full)
 {
-    bool enqueued = false;
-    for (int client = 0; client < config.num_clients; client++) {
+    for (int client = 0; client < config.num_clients && !*drv_full; client++) {
```
Courtney3141: you don't set `*drv_full` before using it.

examples/firewall/net_components/firewall_network_virt_tx.c
```diff
@@ -48,13 +52,19 @@ static int extract_offset_fw_client(uintptr_t *phys)
     return -1;
 }
 
-static void tx_provide(void)
+/* sdfgen sizes tx_queue_drv for the net clients only, not the firewall
+clients also feeding it, so fullness must be checked before every enqueue. */
+static void tx_provide_net_clients(bool *enqueued, bool *drv_full)
```
Courtney3141: I think it's a better idea to return `enqueued` rather than use a pointer. Firstly as you may as well return something, and second currently you are using it un-initialised.

examples/firewall/net_components/firewall_network_virt_tx.c
```diff
@@ -48,13 +52,19 @@ static int extract_offset_fw_client(uintptr_t *phys)
...
 {
-    bool enqueued = false;
-    for (int client = 0; client < config.num_clients; client++) {
+    for (int client = 0; client < config.num_clients && !*drv_full; client++) {
         bool reprocess = true;
         while (reprocess) {
             while (!net_queue_empty_active(&tx_queue_clients[client])) {
+                if (net_queue_full_active(&tx_queue_drv)) {
```
Courtney3141: We will need to restructure this code a little, as it is currently not participating in the networking "signalling protocol" (the communication protocol used between network clients and the virtualiser to avoid deadlocks). I can help with this later :)

examples/firewall/net_components/firewall_network_virt_tx.c
```diff
@@ -105,9 +129,24 @@ static void tx_provide(void)
...
+}
+
+static void tx_provide(void)
+{
+    bool enqueued = false;
+    bool drv_full = false;
+
+    if (fw_clients_priority) {
+        tx_provide_fw_clients(&enqueued, &drv_full);
+        tx_provide_net_clients(&enqueued, &drv_full);
```
Courtney3141: If `drv_full == true`, why would you even call the next function?

Courtney3141 (changes requested):
Your changes to routing.c do fix the issue, but there is still a slight problem with the virtualiser solution. 

Your solution *does* check if the queue is full, but we actually need something stronger than that. We need to check how many packets we have already sent to the driver, without them being freed (+1 each time a packet is enqueued in active, -1 each time a packet is dequeued from free). 

This is because we don't want to have to change the driver code for now, and the driver code *also* assumes that queues cannot be over-filled. Since we can't add fullness checks to the driver, the virtualiser needs to ensure this on behalf of the driver. It just needs to ensure that the total outgoing packets always remains <= the capacity of the driver's queues.

So rather than checking if queues are full, you will need to maintain a count of the number of outgoing packets, and stop enqueueing once it has reached capacity.

This is also done in the new [virtual switch component](https://github.com/au-ts/sddf/blob/main/network/components/vswitch.c#L109).

yumnazaheed (commented):
- firewall_network_virt_tx.c: swapped the fullness check for a count (num_queued_to_driver) since the driver can't check itself. It increments on send, and decrements on free
- round robin every net and firewall client one buffer at a time, alternating which group and which client goes first each pass, so nothing gets starved
- routing.c: moved the drop logic into transmit_packet() itself, so callers don't need to check anything

Issue #339, opened 2026-08-14
Build System Issues with libc


midnightveil:
@0aids constantly runs into this. There seems to be issues when compiling lionsos with make on some systems where it will often build things in a way that breaks the build, so you need to get lucky building several times for it to not break.

We really need to fix up our build system.

wom-bat:
Needs more details.  What breaks?  How does it break?

PR #345 fixes one obvious issue, but this report has too little information for me to see if that's the problem being reported.

0aids:
I think these issues were just skill issues on my end (forgot to state certain dependencies for build targets).
There was another issue where i think make's `$^` variable was injecting extra flags from somewhere, but I can no longer reproduce this issue. If I can reproduce this issue then I will reopen this issue.

dumsum:
I think we should keep this open, though yes it was originally vague.

> I think these issues were just skill issues on my end (forgot to state certain dependencies for build targets).

The point is it somehow needs to be easy to get this right. You're not the first to miss this kind of thing. I suspect what's missing is some kind of a build system "design" - but I'm not exactly sure what that looks like.

I've tried to address some issues in the past but the "solutions" often create other problems. I'm happy to pick this up ongoing but would be good to have specific examples consolidated in this issue.

Off the top of my head:
- Parallel / multithreaded builds are still unreliable despite repeated attempts to specify all dependencies correctly. Anecdotally this is commonly `libc`, usually the headers dependency: looks like kitty still has this issue for example: https://github.com/au-ts/lionsos/blob/f7f02f4fb9afc6ce6c17cee99a6c6f69b4775d97/examples/kitty/kitty.mk#L154
- `lib_sddf_lwip` and `libmicrokitco` "variants" are a bit janky imo, at least with the approach of specifying differently suffixed `CFLAGS`: `LIB_SDDF_LWIP_CFLAGS_0` and `LIB_SDDF_LWIP_CFLAGS_1` to build `lib_sddf_lwip_0.a` and `lib_sddf_lwip_1.a` respectively.
 - snippets including their own dependencies caused problems in the past. https://github.com/au-ts/lionsos/commit/16ca6fa853dc817493bd1abe675af416a30afb68 was an attempted fix, but the problem now is the example needs to know about the dependencies of the snippet (and can create strict snippet inclusion order requirements).
[...]

PR #345, opened 2026-09-19
Make ethernet driver depend on libc.a

wom-bat:
The make snippet building the virtio ethernet driver for qemu doesn't know it needs to have libc built first.  The IMAGES variable contains a different name for the driver from the one it is first built as.

Adding a dependency for libc.a explicitly for the Ethernet Driver elf fixes that,


wom-bat:
fixes bug #339

midnightveil (approved):

